PR #62, opened 2024-06-06
fix: rename duplicate prefix argument to unreal-plugin-directory

joel-wong-aws:
When reviewing the Deadline Cloud submitter installer, I noticed that there was two arguments with the `--prefix` CLI option

This looks like a copy-paste error for the Unreal Engine component of the installer. The current behaviour when the `--prefix` CLI argument is used is unclear - it's surprising that the installer builds at all. For me, the installer errors out when the argument is used.

Also swapped the description and explanation fields for the parameter as the description should be the short-form and the explanation should be the long form.

Customers are able to use the `--prefix` argument when using the installer with Unreal Engine.

#### Before
`DeadlineCloudSubmitter-windows-x64-installer.exe --help`

image

`DeadlineCloudSubmitter-windows-x64-installer.exe --prefix C:\Users\test-user\installbuildertest --mode unattended --unattendedmodeui minimal --enable-components deadline_cloud_for_unreal_engine` 

image

image

#### After

`DeadlineCloudSubmitter-windows-x64-installer.exe --help`

image

`DeadlineCloudSubmitter-windows-x64-installer.exe --prefix C:\Users\test-user\installbuildertest --mode unattended --unattendedmodeui minimal --enable-components deadline_cloud_for_unreal_engine --unreal-plugin-directory C:\Users\test-user\installbuildertest`

<successful installation!>

image

No, not required

No, as this CLI option never worked previously, this is not considered a breaking change.

----

### DIFF
--- a/install_builder/deadline-cloud-for-unreal-engine.xml
+++ b/install_builder/deadline-cloud-for-unreal-engine.xml
@@ -56,13 +56,14 @@
 		</stringParameter>
         <directoryParameter>
             <name>unreal_plugin_dir</name>
-            <description>Path to the Deadline Cloud Unreal Plugin directory</description>
-            <explanation>Deadline Cloud Unreal Plugin directory</explanation>
+            <description>Deadline Cloud Unreal Plugin directory</description>
+            <explanation>Path to the Deadline Cloud Unreal Plugin directory</explanation>
             <value></value>
             <default>C:\Program Files\Epic Games\UE_5.2\Engine\Plugins\UnrealDeadlineCloudService</default>
             <allowEmptyValue>0</allowEmptyValue>
             <ask>yes</ask>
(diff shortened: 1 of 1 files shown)


crowecawcaw (approved):
Just to double check, the `unreal-plugin-directory` option was already wired up and functional but this option just had the wrong name?

epmog (approved):

joel-wong-aws:
> Just to double check, the `unreal-plugin-directory` option was already wired up and functional but this option just had the wrong name?

That's correct. When using the CLI to install the Deadline Cloud submitters, this argument was actually unusable. It duplicated the `prefix` argument that is used to choose the installation directory for the Deadline Cloud CLI python files,m and the Deadline Cloud CLI `prefix` argument took precedence.

This doesn't modify the UI flow at all except for swapping the description and explanation.

PR #63, opened 2024-06-07
added Unreal Engine compatibility requirements to README. Customers were asking/compiling against 4.X versions  - the internal docs say that the code supports 5.2.1 upwards.

andymanaws:
…ted in the lower docs, but customers have been trying to build on 4.X

Cutomers were confused about which versions are supported.

The internal build docs stipulate 5.2.1 - I am just bubbling this up to the top.

Just elevation of min requirements.

Its just a README change, based on deeper internal docs that you have to build to see.

NO

----

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,8 +17,9 @@ AWS Deadline Cloud for Unreal Engine is a python package that allows users to cr
 
 This library requires:
 
-1. Python 3.9 or higher; and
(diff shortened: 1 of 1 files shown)


baxeaz:
I've added this change as part of https://github.com/aws-deadline/deadline-cloud-for-unreal-engine/commit/e1692b22b8d49bd7fa3eb82b676abd2de2e8db49 - thanks!

PR #72, opened 2024-10-04
feat: Submitter installer improvements

baxeaz:
Submitter installer defaulted to trying to install in Unreal version 5.2's install folder, even when users did not have 5.2 but did have another version installed.

Submitter installer forced customers to have created the UnrealDeadlineCloudService plugin folder before installing.

Several places in did not correctly reference 5.2 as the minimum version.

Updating default installed version path of unreal to 5.4 (Latest official release)

Changed some comments to reflect support for 5.2 and above.

Added a registry check for most recently installed version which will override default when found.

Removed requirement for 'UnrealDeadlineCloudService' Plugins folder to exist for installation to work, only the engine plugins folder needs to exist.

Better customer experience installing unreal submitter.

Built and ran submitter installer with a number of different version combinations set in the registry.

Validated UnrealDeadlineCloudService folder was created when necessary and did not block installation if not present.

Ran on Linux to verify no unexpected interactions/errors due to changes.
No

No
----

### DIFF
--- a/install_builder/deadline-cloud-for-unreal-engine.xml
+++ b/install_builder/deadline-cloud-for-unreal-engine.xml
@@ -41,6 +41,34 @@
         <actionList>
             <setInstallerVariable name="all_components" value="${all_components} deadline_cloud_for_unreal_engine"/>
             <setInstallerVariable name="unreal_deps_platform" value="windows"/>
+            <setInstallerVariable name="unreal_install_path" value="C:\Program Files\Epic Games\UE_5.4"/>
+            <registryFind>
+                <keyPattern>*5.*</keyPattern>
+                <namePattern>InstalledDirectory</namePattern>
+                <findAll>1</findAll>
+                <rootKey>HKEY_LOCAL_MACHINE\SOFTWARE\EpicGames\Unreal Engine</rootKey>
+                <searchDepth>1</searchDepth>
+                <variable>unreal_registry_list</variable>
+            </registryFind>
+            <foreach>
+                <variables>key name value</variables>
+                <values>${unreal_registry_list}</values>
+                <actionList>
+                    <actionGroup>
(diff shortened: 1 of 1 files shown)


crowecawcaw (approved):

install_builder/deadline-cloud-for-unreal-engine.xml
```diff
@@ -41,6 +41,34 @@
...
+                <keyPattern>*5.*</keyPattern>
+                <namePattern>InstalledDirectory</namePattern>
+                <findAll>1</findAll>
+                <rootKey>HKEY_LOCAL_MACHINE\SOFTWARE\EpicGames\Unreal Engine</rootKey>
+                <searchDepth>1</searchDepth>
+                <variable>unreal_registry_list</variable>
+            </registryFind>
+            <foreach>
+                <variables>key name value</variables>
+                <values>${unreal_registry_list}</values>
```
epmog, line 55: So to clarify, each unreal installation (major.minor?) will list its installation directory here, and we're just presenting the last one that we find to the user? They can still override and choose a different one?

Additionally, how are those registry values sorted? Latest will always be last?

epmog (commented):
Just a few questions, otherwise it looks good

install_builder/deadline-cloud-for-unreal-engine.xml
baxeaz: Yep!  Each version goes into the registry here, most recently installed version last, and what we're gathering here is just the suggested default which we still present them with the option to override with some other folder, as long as they point to a plugins folder which exists.

epmog: perfect, thanks!

epmog (approved):

PR #78, opened 2024-10-28
feat: split MRQ Job level sequence into shot chunks

alf-devnull:
We want to make rendering more flexible: add the ability to divide shots into chunks and run Unreal Engine once with additional CMD arguments to render all of these chunks within one running session

Add new parameters ChunkSize and ChunkId into Render OpenJD Step which will define what bunch of shots enabled by Artist in the MRQ Job will be rendered in the single Task.
For example, LevelSequence has 20 shots, 15 - enabled. Artist set ChunkSize as 4.
Therefore **ChunkSize=4**, **ChunkId=[0, 1, 2, 3]** (ChunkId 0 has shots 1,2,3,4, ChunkId 1 - 5,6,7,8, ChunkId 2 - 9,10,11,12, Chunk 3 - 13,14,15)

Launch `unreal-engine-openjd` with arguments `daemon start` in OpenJD Environment `onEnter` action and stop it with arguments `daemon stop` in `onExit` action. Run `unreal-engine-openjd`  with arguments `daemon run` in OpenJD "Render" Step script.

Add new parameter `ExtraCmdArgs` into OpenJD Job which will define additional arguments to run Unreal Editor with. Pass them to new UnrealEditor-Cmd subprocess in `unreal-engine-openjd`

It will be possible to render only part of the shots, which will reduce the total waiting time for the desired result.
Unreal Engine will be launched once on each machine before all tasks. This will reduce the overall rendering time

Unit tests was written for Shot Chunks feature that include different edge cases
QA engineers performed E2E testing

This chage was described in appropriate Technical Design Document

**No**
----


test/deadline_submitter_for_unreal/unit/cases/test_shot_chunks.py
```diff
@@ -0,0 +1,52 @@
...
+        chunk_size_param = next((p for p in parameters if p['name'] == 'ChunkSize'), None)
+        assert chunk_size_param is not None
+        assert chunk_size_param['range'][0] == chunk_size
+
+        chunk_id_param = next((p for p in parameters if p['name'] == 'ChunkId'), None)
+        assert chunk_id_param is not None
+        assert chunk_id_param['range'] == expected_chunk_id
+
+    def test_chunk_size(self):
+        # TODO find better approach on parametrizing unittest.TestCase cases
```
baxeaz: IMO this is one of the reasons to prefer pytest over unittest - I know we've got a bunch of unittest.TestCases around in this repo already though.  I've been holding off on reviewing since I know you were looking at at least a few more changes from our offline discussion - is there more test coverage coming for these changes though, was that pending this TODO getting sorted out?

alf-devnull: We're not planning yet on increasing the test coverage of the Python side for the recent changes since other changes affect YAML templates and adaptor side that already covered with tests.

src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
```diff
@@ -280,6 +280,24 @@ def _start_unreal_client(self) -> None:
 
         unreal_exe = "UnrealEditor-Cmd"
         unreal_project_path = self.init_data.get("project_path", "")
+        extra_cmd_str = self.init_data.get("extra_cmd_args", "")
```
baxeaz, line 283: For all of this for example we feel like we've got covered in tests elsewhere?

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
```diff
@@ -187,6 +187,31 @@ def create_queue_from_job_args(
         name = job_name or Path(level_sequence_path).stem
         render_job.job_name = name
 
+    @staticmethod
+    def create_queue_from_queue_asset(movie_pipeline_queue_subsystem, movie_pipeline_queue_asset_path: str):
```
baxeaz: Is this for a future PR, not seeing it used anywhere currently. And does/will it need test coverage?

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
```diff
@@ -187,6 +187,31 @@ def create_queue_from_job_args(
...
+        movie_pipeline_queue_asset = unreal.EditorAssetLibrary.load_asset(movie_pipeline_queue_asset_path)
+        pipeline_queue.copy_from(movie_pipeline_queue_asset)
+
+    @staticmethod
+    def enable_shots_by_chunk(
+            render_job,
+            task_chunk_size: int,
+            task_chunk_id: int
+    ):
+        all_shots_to_render = [shot for shot in render_job.shot_info if shot.enabled]
```
baxeaz, line 208: Same question here and the run_script call below that calls this, I see we've got a basic test for UnrealCustomStepHandler.run_script but is this covered?  And if the methods are called in tests somewhere are we validating all of this new behavior as well?

src/deadline/unreal_submitter/unreal_open_job/open_job_description.py
```diff
@@ -413,3 +452,69 @@ def _save_manifest_file(self, mrq_job):
...
+        in_process_executor_settings = unreal.get_default_object(unreal.MoviePipelineInProcessExecutorSettings)
+
+        # Append all of inherited command line arguments from the editor
+        inherited_cmds: str = in_process_executor_settings.inherited_command_line_arguments
+
+        # Sanitize the commandline by removing any execcmds that may have passed through the commandline.
+        # We remove the execcmds because, in some cases, users may execute a script that is local to their editor build
+        # for some automated workflow but this is not ideal on the farm.
+        # We will expect all custom startup commands for rendering to go through the `Start Command` in the MRQ settings
+        inherited_cmds = re.sub(pattern='(-execcmds="[^"]*")', repl="", string=inherited_cmds)
```
baxeaz, line 472: Same question about coverage here

src/deadline/unreal_submitter/unreal_open_job/open_job_description.py
```diff
@@ -359,11 +386,23 @@ def _build_steps(self, mrq_job) -> list[JobStep]:
         preset_overrides: unreal.DeadlineCloudJobPresetStruct = mrq_job.preset_overrides
         unreal.log(f"Preset overrides: {preset_overrides}")
 
+        shots_to_render = []
+        for shot_index, shot in enumerate(mrq_job.shot_info):
```
baxeaz: And for this I see we've got some general tests of OpenJobDescription in test_unreal_open_job but are we testing this part out somewhere?

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
Iskander7d: Yes, this for future PR. Will be covered later, not used right now

alf-devnull: Yes, its for future PR. Will be covered later

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
alf-devnull: Will add test case for that on the adaptor side

src/deadline/unreal_submitter/unreal_open_job/open_job_description.py
alf-devnull: Will add test case for building cmd args string

src/deadline/unreal_submitter/unreal_open_job/open_job_description.py
alf-devnull: Yes, we check the shot chunks building in the OpenJob Step enity. But its a good idea to check this data before passing to the Step

Iskander7d: test will be added

test/deadline_adaptor_for_unreal/unit/UnrealClient/step_handlers/test_render_step_handler.py
```diff
@@ -0,0 +1,71 @@
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+import sys
+import pytest
+from unittest.mock import MagicMock
+
+unreal_mock = MagicMock()
+unreal_mock.log.return_value = MagicMock()
+
+sys.modules['unreal'] = unreal_mock
```
baxeaz: I wonder if it might make more sense to wrap our logging calls in some other method which could either test for importing unreal and fall back on prints or something if it's not around, or we could patch that wrapper instead, or maybe there's some other better solution I'm not thinking of.  Seems like needing to do this all over in our tests isn't the best solution long term.  What do you think?   If we did decide this was the best approach currently we should comment what's going on at least I think

alf-devnull: In most of the submitter we use a lot of unreal objects, methods and classes. The unreal Python module does not contain any logic, only declarations. Therefore, it is not advisable to mock specific objects and methods used when converting MRQ jobs to OpenJob. In this case, we test methods that do not rely on unreal objects (we managed to get by with two mocked classes MRQJob and ShotInfo), and logging is the only thing used there. Therefore, this is more of a single case than a reason to come up with a solution for logging in all tests (unreal.log is the most trivial thing that can be there)

baxeaz: Ok - we've got a couple cases of it happening in this CR so figured it would be worth considering now, but yes if you think it's unlikely to need to be applied in many other cases maybe that's not such an issue.  For the logging case at least it seems like we should be able to install a customer logger when the unreal module is available and then just route all of our logging through python logging - but yeah if we do think this is a one time issue maybe that's outside the scope of what we're really doing here.

src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
```diff
@@ -291,9 +309,23 @@ def _start_unreal_client(self) -> None:
...
         ]
 
+        extra_cmd_args = extra_cmd_str.split(" ")
+
         args = [unreal_exe, unreal_project_path]
         args.extend(log_args)
-        args.append(f"-execcmds=r.HLOD 0,py {client_path}")
+        args.extend(extra_cmd_args)
+        args = [arg for arg in args if arg]  # Remove empty strings
+        args = list(dict.fromkeys(args))     # Remove duplicates
```
erico-aws: Should we use a Set here?

test/deadline_submitter_for_unreal/unit/cases/test_shot_chunks.py
```diff
@@ -0,0 +1,89 @@
...
+            # GIVEN
+            enabled_shots = [ShotInfoMock(enabled=True, outer_name=f'Enabled{i}') for i in range(enabled_shots_count)]
+            disabled_shots = [ShotInfoMock(enabled=False, outer_name=f'Disabled{i}') for i in range(shots_count - enabled_shots_count)]
+            render_job_mock = RenderJobMock(job_name='MockedMrqJob', shot_info=enabled_shots + disabled_shots)
+
+            # WHEN
+            enabled_shot_names = OpenJobDescription.get_enabled_shot_names(render_job_mock)
+
+            # THEN
+            assert len([shot for shot in enabled_shot_names]) == enabled_shots_count
```
erico-aws: Is this not just `len(enabled_shot_names)`?

src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
alf-devnull: you mean use list(set(args) instead of list(dict.fromkeys(args))?

alf-devnull: actually we want to keep the order of keys in original args list:
1. Unreal Executable
2. UProject path
3. Launch args ...
And dict.fromkeys allows to do that

test/deadline_submitter_for_unreal/unit/cases/test_shot_chunks.py
alf-devnull: yes, will fix it

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
```diff
@@ -132,6 +132,11 @@ def executor_failed_callback(executor, pipeline, is_fatal, error):
     def executor_finished_callback(movie_pipeline=None, results=None):
         unreal.log("Render Executor: Rendering is complete")
 
+    @staticmethod
+    def log(message: str):
+        if unreal:
```
baxeaz: This will get us there!  I was kind of imagining we'd subclass logging.Handler and override emit to unreal.log, and only attach that handler when unreal was available (Like in init_unreal).  Just so I understand, was there a reason to prefer this way?

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
```diff
@@ -132,6 +132,11 @@ def executor_failed_callback(executor, pipeline, is_fatal, error):
     def executor_finished_callback(movie_pipeline=None, results=None):
         unreal.log("Render Executor: Rendering is complete")
```
baxeaz: Is there reason some still go through the direct call?

src/deadline/unreal_submitter/unreal_logger/handlers.py
```diff
@@ -0,0 +1,28 @@
+import logging
+
+
+class UnrealLogHandler(logging.Handler):
```
baxeaz: Oh, you did do what I was describing above!  Are there a couple different solutions now though?  Did we need both?

src/deadline/unreal_submitter/unreal_logger/logger.py
```diff
@@ -0,0 +1,33 @@
...
+    LOGGER_CONFIG = json.load(f)
+
+
+def get_logger() -> logging.Logger:
+
+    logging.config.dictConfig(LOGGER_CONFIG)
+
+    unreal_logger = logging.getLogger("unreal_logger")
+
+    unreal_log_handler: Optional[UnrealLogHandler] = next(
```
baxeaz: What are we doing in here?  We'll want to comment this at least if it is all necessary

src/deadline/unreal_submitter/unreal_logger/logger.py
```diff
@@ -0,0 +1,33 @@
...
+
+with open(f"{os.path.dirname(__file__)}/config.json", "r") as f:
+    LOGGER_CONFIG = json.load(f)
+
+
+def get_logger() -> logging.Logger:
+
+    logging.config.dictConfig(LOGGER_CONFIG)
+
+    unreal_logger = logging.getLogger("unreal_logger")
```
baxeaz: We never actually add our UnrealLogHandler with [addHandler](https://docs.python.org/3/library/logging.html#logging.Logger.addHandler) it looks like?  Again maybe there's a good reason, but it seems like we're doing more work to manage the handler than I'd expected we'd need to..

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
alf-devnull: the reason is where unreal logger wrapper lives. I think its not a good idea to create it in src/deadline to make available in adaptor and submitter since only in single render handler (inside adaptor) unreal logic used. And dont want to duplicate code in adaptor to implement the same logger logic. I think we want to follow the approach where only unreal_adaptor and unreal_submitter as independent packages live inside deadline folder

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
alf-devnull: Its ok, because this callback is needed just for telling the adaptor that unreal client finished the task, no need to wrap with our static log method

src/deadline/unreal_submitter/unreal_logger/handlers.py
alf-devnull: What do you mean?

src/deadline/unreal_submitter/unreal_logger/logger.py
alf-devnull: Since unreal can handle both stdout and unreal.log messages we can get log duplication for each logger call. So as default behaviour - we remove unreal handler if its not initialized properly and use default stdout. And vice versa, we delete the regular one and leave the unreal one

src/deadline/unreal_submitter/unreal_logger/logger.py
alf-devnull: we setup our custom "unreal_logger" logger with config.json file that lives inside same directory and set up it with `logging.config.dictConfig`. Here we describe the formatters, handlers and the logger itself. I think its better to keep configuration in the appropriate file instead of configuring in the code. What do you think?

src/deadline/unreal_submitter/unreal_logger/handlers.py
baxeaz: We have both UnrealLogHandler handling optional unreal logging in one way and UnrealRenderStepHandler doing using that other approach of if unreal:  unreal.log - I was expecting everything to use the same approach

alf-devnull: Ok, in that way we could try to add logger creation in the top of render_step_handler depending on unreal import result

src/deadline/unreal_submitter/unreal_logger/logger.py
baxeaz: Ahh I think I see - our config file has both handlers declared.  Yeah the other options would be to have those live in different files that we used depending on which we were trying to initialize.  Or yeah I guess the question of data vs code - code might make that a little easier to just add the one thing we want.  It just seems a bit odd to be adding two loggers then conditionally removing one rather than just add the one we want.  But yes if we do keep things this way we'll want to comment what's going on because it looks a bit unusual

src/deadline/unreal_submitter/unreal_logger/logger.py
baxeaz: I was more surprised we didn't add our handler to the logger with addHandler, then just go through the normal python logger.info type methods.  Commented above about this, but I think this is also why we're ending up doing the "add 2 log handlers, remove 1" steps currently, so yeah it might look cleaner if we did it as code and only added one, or had two data files potentially.

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
baxeaz: Definitely doesn't make sense to duplicate code if we can avoid it, I agree!  But having shared code "somewhere" I don't think would be a bad thing.  Consistency across the codebase I think would be valuable long term. It's not a blocking issue to have it this way though

src/deadline/unreal_submitter/unreal_logger/handlers.py
baxeaz: That we be great if we could be more consistent without too much trouble!

src/deadline/unreal_logger/logger.py
```diff
@@ -0,0 +1,37 @@
...
+
+def get_logger() -> logging.Logger:
+    """
+    Returns an instance of logging.Handler.
+    Attach handler :class:`deadline.unreal_logger.handlers.UnrealLogHandler` if unreal module is
+    available
+    """
+
+    unreal_logger = logging.getLogger("unreal_logger")
+    print(id(unreal_logger))
```
baxeaz: Left over from debugging?

alf-devnull: oh, yes, sorry, will remove it right now

src/deadline/unreal_logger/logger.py
```diff
@@ -0,0 +1,37 @@
...
+    except ModuleNotFoundError:
+        unreal = None
+
+    # can be called outside UE so need to check before adding UE specific handler
+    if unreal is not None:
+        unreal_log_handler = UnrealLogHandler(unreal)
+        unreal_log_handler.setLevel(logging.DEBUG)
+
+        # If we found existed UnrealLogHandler, skip adding duplicate
+        existed_unreal_log_handler = next(
```
baxeaz: Much cleaner now!  Only thing I'm wondering is whether this could all just basically be an "initialized" bool we set after we've done it once, or is there something more complex we're handling here?

alf-devnull: you mean using global `initialized` bool in try except block and check for it instead of `unreal is not None`?

alf-devnull: Theoretically, the module unreal is either available or not (run in session or outside of it). Therefore, we can move the import to the top of the file and set the variable. But I still think the check for the presence of the handler should be left

Smth like this:
```
try:
    import unreal  # noqa: F401
    unreal_initialized = True
except ModuleNotFoundError:
    unreal = None
    unreal_initialized = False


def get_logger() -> logging.Logger:
    unreal_logger = logging.getLogger("unreal_logger")
    unreal_logger.setLevel(logging.DEBUG)
    
    # can be called outside UE so need to check before adding UE specific handler
    if unreal_initialized:
        ...
```

baxeaz: Well once we've called get_logger once and either added the handler or not added the handler we should be able to use the bool to know if we need to potentially add it again I think?

alf-devnull: Good suggestion, but we still need to know the first time if there is a suitable unreal handler. So iteration with next is unavoidable, I think

alf-devnull: So lets do this way for example
```
try:
    import unreal  # noqa: F401

    UNREAL_INITIALIZED = True
except ModuleNotFoundError:
    unreal = None
    UNREAL_INITIALIZED = False


UNREAL_HANDLER_ADDED = False

def get_logger() -> logging.Logger:
    unreal_logger = logging.getLogger("unreal_logger")
    unreal_logger.setLevel(logging.DEBUG)

    global UNREAL_HANDLER_ADDED

    # can be called outside UE so need to check before adding UE specific handler
    if not UNREAL_HANDLER_ADDED and UNREAL_INITIALIZED:
       # just add handler
       ...
```

baxeaz (approved):

erico-aws (approved):

PR #82, opened 2024-11-20
feat: Data assets refactoring, support UnrealGameSync

alf-devnull:
We want to make rendering more flexible. It is necessary to give an ability to build the OpenJob with parameters, Steps and Environments from UE Slate and MRQ Job. It's also required to deal with Unreal project files via UnrealGameSync instead of default S3 manager.

Solution consist of next main parts:
1. Refactored UE data assets. Instead of aproach based on predefined OpenJD templates we create data assets for each Job, Step and Environment OpenJD entity. Each of them is built according to the OpenJD specification and give the ability to User to create this entities from prepared in advance YAML templates. In this way, Users can configure jobs and parameter values straight from Editor.
2. Update Python entities according to UE data assets. In this part main business logic of building OpenJob bundle is executed
3. Prepare ready-to-go templates for Jobs using S3 file manager and UGS source control system. These templates contains necessary parameters, scripts and environment variables for Job, Step and Environment.

1. It will be possible to create OpenJob straight from Unreal Editor combining different steps and environments with customized parameters (such as ChunkSize, ExtraCmdArgs, Executable, etc.).
2. Using UGS will speed up the submit, because in fact, instead of uploading User files to S3, it will be replaced by downloading files from the depot to the worker node before all tasks, and we also shift responsibility for version control to UGS, because this is what it is intended for.

Unit tests
QA engineers performed E2E testing

This chage was described in appropriate Technical Design Document

Yes
----


src/unreal_plugin/Content/Python/init_unreal.py
```diff
@@ -3,6 +3,7 @@
 import os
 import sys
 
+
 remote_execution = os.getenv("REMOTE_EXECUTION", "False")
 if remote_execution != "True":
     from deadline.unreal_logger import get_logger
```
baxeaz, line 9: I know this is from a previous PR - I think this needs to move down below the path setup we're doing though, otherwise if you don't have a system deadline installed I don't think we're going to resolve deadline (Ran into this issue in testing).   (Line 19 will need to move down as well then)

baxeaz:
Closed in favor of pull #89 which includes all of these changes

PR #84, opened 2024-11-21
ci: Adding a SonarQube config file to match Unreal Engine development

baxeaz:

Unreal has a number of coding conventions which don't match default C++. Some of these were being flagged by SonarQube as incorrect. This will lead to unnecessary development churn and frustration.

Add a custom sonarqube configuration file to allow Unreal conventions.

Better developer experience.

Testing with an incoming PR.

No

No

----

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -1,0 +1,28 @@
+sonar.projectKey=unreal-plugin
+sonar.projectName=UnrealPlugin
+sonar.sources=.
+sonar.cplusplus.file.suffixes=.cpp,.h,.hpp
+sonar.cpp.file.suffixes=.cpp,.h,.hpp
+
+# Ignore Unreal-specific naming conventions
+sonar.issue.ignore.multicriteria=e1,e2,e3,e4,e5
+
+# Allow keyword virtual on overrides
+sonar.issue.ignore.multicriteria.e1.ruleKey=cpp:S3471
+sonar.issue.ignore.multicriteria.e1.resourceKey=**/*.h
+
+# Allow LOCTEXT_NAMESPACE macro
+sonar.issue.ignore.multicriteria.e2.ruleKey=cpp:S5028
+sonar.issue.ignore.multicriteria.e2.resourceKey=**/*
+
+# Allow Unreal's memory management patterns
+sonar.issue.ignore.multicriteria.e3.ruleKey=cpp:S5025
+sonar.issue.ignore.multicriteria.e3.resourceKey=**/*
+
+# Allow keyword virtual on overrides
+sonar.issue.ignore.multicriteria.e4.ruleKey=cpp:S3471
+sonar.issue.ignore.multicriteria.e4.resourceKey=**/*.h
+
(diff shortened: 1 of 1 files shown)


sonar-project.properties
```diff
@@ -0,0 +1,62 @@
...
+# Exclude build directories and third-party code
+sonar.exclusions=**/Build/**,**/Intermediate/**,**/ThirdParty/**,**/Content/**
+
+# Configure C++ analysis
+sonar.cpp.errorRecoveryEnabled=true
+sonar.cpp.microsoft=true
+sonar.cpp.standardLibrary=microsoft
+
+# Ignore Unreal-specific naming conventions
+sonar.issue.ignore.multicriteria=e1,e2,e3,e4,e5,e6
```
mwiebe: What does this specifically do, how does it relate to naming conventions?

sonar-project.properties
```diff
@@ -0,0 +1,62 @@
...
+
+# Custom parameters for Unreal-specific patterns
+sonar.cplusplus.complexity.threshold=15
+sonar.cplusplus.lines.of.code.per.file.threshold=1000
+sonar.cplusplus.maximum.function.parameters.threshold=8
+
+# Allow UPROPERTY macro usage
```
eherozhao: May I ask why do we choose these rule key? Any guideline we followed?

sonar-project.properties
```diff
@@ -0,0 +1,62 @@
...
+sonar.issue.ignore.multicriteria.e2.resourceKey=**/*.h
+
+# Allow enum "Collisions" for enum BlueprintType declarations
+sonar.issue.ignore.multicriteria.e2.ruleKey=cpp:S841
+sonar.issue.ignore.multicriteria.e2.resourceKey=**/*.h
+
+# Enable specific rules useful for Unreal development
+sonar.cpp.suffixes.headers=.h,.hpp
+sonar.cpp.suffixes.sources=.cpp
+sonar.c.file.suffixes=
```
mwiebe: Could these go in the top block along with `sonar.cpp.file.suffixes=.cpp,.h,.hpp`? They look thematically similar to those.

sonar-project.properties
baxeaz: It's a good call out - I really don't think we need these at all.  Removed this and several other parts I'm not yet sure we need.

sonar-project.properties
baxeaz: Specifically these are just declaring the existence of named ignores we'll use below I believe.  The documentation for this file is not great - a lot of what's here it seems other folks had to just figure out.  SonarQube seems to vastly prefer these are configured through the web UI.  I probably should have asked if we had a policy around this and also preferred setting these in the UI - setting it in a config file didn't seem at all out of the ordinary at first, and likely far more maintainable.

sonar-project.properties
baxeaz: Actually I'm not sure we need both - Trimmed out a bunch, but going to do some experimenting locally, so not looking for another review quite yet

PR #85, opened 2024-11-22
chore(deps): Upgrading required versions of deadline and openjd

baxeaz:
The openjd EntryPoint timeout parameter which is required by the current version mainline version of the plugin is only in openjd 0.8.1 and above.  Openjd 0.8.1 requires pywin32 308, but deadline 0.48.9 required pywin32 306.

Upgrade dependency versions.

Plugin should install without complaints about version conflicts.

hatch run test
hatch run fmt
hatch run lint
hatch build and pip install

No

No

----


joel-wong-aws:
supernit: the commit history on this branch is very complicated maybe you need a rebase?

joel-wong-aws (approved):

epmog (approved):

PR #93, opened 2024-12-04
fix: Submitter installer fixes

baxeaz:
Fixes intended to support an intermediate release of the Unreal integration while we continue to take in the larger new features currently in flight.

Importing logger after setting uninstall files in submitter installer.  Adding logger to submitter installer.  Setting uninstall files.

- The new logger module was being imported before adding submitter installer installed dependency path to sys.path
- The new logger module wasn't being included in submitter installer.
- Submitter installer didn't properly uninstall the unreal plugin.

- Move the code where we import logger further down.
- Add logger to submitter installer
- Add the unreal plugin to our uninstall files.

Unreal submitter should function properly when installed from the submitter installer and uninstall properly.

Built, installed and manual testing with submitter installer.
hatch build/hatch run lint/hatch run fmt/hatch run all:test
Uninstalled program successfully

No

No
----

### DIFF
--- a/install_builder/deadline-cloud-for-unreal-engine.xml
+++ b/install_builder/deadline-cloud-for-unreal-engine.xml
@@ -18,6 +18,17 @@
             </distributionFileList>
         </folder>
         <folder>
+            <description>Logger Files</description>
+            <destination>${unreal_plugin_dir}/Content/Python/libraries/deadline/unreal_logger</destination>
+            <name>unreal_logger</name>
+            <platforms>all</platforms>
+            <distributionFileList>
+                <distributionDirectory allowWildcards="1">
+                    <origin>components/deadline-cloud-for-unreal-engine/src/deadline/unreal_logger/*</origin>∂
+                </distributionDirectory>
+            </distributionFileList>
+        </folder>
+        <folder>
             <description>Dependency Files</description>
             <destination>${unreal_plugin_dir}/tmp/unreal_deps</destination>
             <name>unreal_deps</name>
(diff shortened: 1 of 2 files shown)


install_builder/deadline-cloud-for-unreal-engine.xml
```diff
@@ -18,24 +18,24 @@
...
-            <description>Prebuild Outputs</description>
-            <destination>${unreal_plugin_dir}</destination>
-            <name>unreal_prebuilds</name>
+            <description>Dependency Files</description>
+            <destination>${unreal_plugin_dir}/tmp/unreal_deps</destination>
+            <name>unreal_deps</name>
             <platforms>all</platforms>
             <distributionFileList>
                 <distributionDirectory allowWildcards="1">
-                    <origin>components/deadline-cloud-for-unreal-engine/UnrealDeadlineCloudService/*</origin>
```
erico-aws, line 38: Do we not need the files from here anymore?

baxeaz: No we're currently not using this so I wanted to remove it until/unless we are.  The prebuilds are keyed to the Unreal version so for each new patch version of Unreal that's released (Which they may do up to weekly after releasing a new minor version release like they just did with 5.5) we'll often be out of date before our next submitter installer is even released currently.  We're investigating other options for this.

erico-aws (approved):

joel-wong-aws (approved):

PR #96, opened 2024-12-06
feat!: ExtraCmdArgsFile parameter to handle param length limit

alf-devnull:
Unreal Submitter collect necessary system command line arguments + user's custom argumets and pass them as STRING  parameter. The result string can exceed the limit of OpenJD STRING parameter length limit of 1024 characters.

On Submitter:
Add new parameter ExtraCmdArgsFile of type PATH where the result arguments string saved to the Job template.

On Adaptor:
Read from file ExtraCmdArgsFile

User can submit the job with UE cmd arguments of any length

Unittest + QA

Yes

No
----

### DIFF
--- a/src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
+++ b/src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
@@ -289,8 +289,6 @@ class UnrealAdaptor(Adaptor[AdaptorConfiguration]):
             with open(extra_cmd_args_file, "r") as f:
                 extra_cmd_str = f.read()
 
-        print(f"EXTRA CMD ARGS: {extra_cmd_str}")
-
         # Everythiing between -execcmds=" and " is the value we want to keep
         match = re.search(r'-execcmds=["\']([^"\']*)["\']', extra_cmd_str)
         if match:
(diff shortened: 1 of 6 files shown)


baxeaz (approved):

src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
```diff
@@ -280,7 +280,14 @@ def _start_unreal_client(self) -> None:
 
         unreal_exe = "UnrealEditor-Cmd"
         unreal_project_path = self.init_data.get("project_path", "")
-        extra_cmd_str = self.init_data.get("extra_cmd_args", "")
+
+        # First, read args from file since it can be too long to pass
+        # them to Job parameter (1024 chars limit)
+        extra_cmd_str = ""
+        extra_cmd_args_file = self.init_data.get("extra_cmd_args_file", "")
```
karthikbekalp, line 287: Will this change break backwards compatibility? 

Wouldn't customers that currently use `extra_cmd_args` field now fail to work?
Or did it never work as the command string were always greater than 1024 characters?

karthikbekalp (commented):
The changes look good to me but I wonder if the changes here are backwards incompatible? If so should we mark the commit message as `feat!` instead?

src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
alf-devnull: This change shouldn't break backwards compatibility because this parameter suppose to be optional and added if only any parameters passed
Also we remove `extra_cmd_args` and use only `extra_cmd_args_file` that always include inherited cmd args from User's Unreal session

leon-li-inspire (approved):

PR #101, opened 2024-12-26
feat!: support Perforce

alf-devnull:
We want to make rendering more flexible. It is required to deal with Unreal project files via Perforce instead of default S3 manager.

- Add new Python and C++ classes that implements P4 Jobs, Steps and Environments
- Add new package `unreal_perforce_utils` with useful P4 functionality to create, sync and delete workspaces

Artsts now have an ability to submit render jobs using Perforce to transfer project files

- unit tests
- QA

Yes

Yes

Yes
----


docs/source/docs_usage/use_cases/create_p4_render_job.rst
```diff
@@ -0,0 +1,89 @@
...
+***********************************
+
+#. Select "Deadline Cloud Perforce Environment". Name an asset, for example "P4SyncCmfEnvironment" (SyncSmf) and open it for editing
+
+   .. image:: /images/create_p4_render_job_0.png
+
+#. Select **Content/Python/openjd_templates/p4/p4_sync_cmf_environment.yml** (**p4_sync_smf_environment.yml**)
+#. Environment variables and Name from YAML will be loaded to data asset
+
+   .. image:: /images/create_p4_render_job_1.png
```
baxeaz, line 43: Are some of these images missing from the PR?  Looks like we only have create_p4_render_job_0.png

docs/source/docs_usage/use_cases/create_p4_render_job.rst
```diff
@@ -0,0 +1,89 @@
+Create P4 Render Job
+======================
+
+To submit MRQ Job from Unreal Project under Perforce repository you need to meet next requirements:
```
baxeaz: "To submit an MRQ Job from an Unreal project in a Perforce repository you need to meet the following requirements:"

docs/source/docs_usage/use_cases/create_p4_render_job.rst
```diff
@@ -0,0 +1,89 @@
...
+   1. Unreal Project should lay under the P4 workspace
```
baxeaz: The Unreal project should be within the P4 workspace

docs/source/docs_usage/use_cases/create_p4_render_job.rst
```diff
@@ -0,0 +1,89 @@
...
+   #. Check that your current P4 connection is established and you’re logged in P4
+
+P4 Render Job structure is pretty same as Default Render Job but contains extra Environments
```
baxeaz: "The P4 Render Job structure is pretty much the same as the Default Render Job but contains an extra Environment" 

- It's just one extra Environment I think right?  The default one now has the Launch environment as well?

src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
```diff
@@ -495,14 +508,15 @@ def on_run(self, run_data: dict) -> None:
             #  This is always an error case because the Unreal Client should still be running and
```
baxeaz: Is this comment no longer correct?  And if so what changed?  It seems like now as long as we had a clean exit we're no longer treating it as an error?

src/deadline/unreal_adaptor/UnrealClient/unreal_client.py
```diff
@@ -24,19 +24,29 @@
 
 logger = get_logger()
 
+# Global variables for keep UnrealClient running and prevent deletion by Unreal garbage collector
```
baxeaz: "# Global variables for keeping UnrealClient running and preventing deletion by the Unreal garbage collector"

src/deadline/unreal_adaptor/UnrealClient/unreal_client.py
```diff
@@ -53,36 +63,55 @@ def close(self, args: Optional[dict] = None) -> None:
...
+                if action is not None:
+                    print(
+                        f"Performing action: {action}",
+                        flush=True,
+                    )
+                    self._perform_action(action)
+            else:  # Any other status or reason
                 print(
-                    f"Performing action: {action}",
+                    f"ERROR: An error was raised when trying to connect to the server: {status} "
```
baxeaz: Is action always empty in this case and so not useful to log here?

src/deadline/unreal_perforce_utils/app.py
```diff
@@ -0,0 +1,354 @@
...
+    logger.info(
+        f"Creating perforce workspace from template: \n"
+        f"Specification template: {specification_template}\n"
+        f"Project: {project_name}\n"
+        f"Overridden workspace root: {overridden_workspace_root}"
+    )
+
+    workspace_name = get_workspace_name(project_name=project_name)
+
+    specification_template_str = json.dumps(specification_template)
```
baxeaz: So it looks like our routine is always that we load the file into JSON in get_workspace_specification_template_from_file, then we call create_perforce_workspace_from_template with the dict which we first turn back to a string to do the replacement, then turn it back to a dict, would it make more sense to just keep things as a string until the replacement is done?

baxeaz: Also might be worth in a comment somewhere here showing what this json roughly looks like

src/deadline/unreal_perforce_utils/app.py
```diff
@@ -0,0 +1,354 @@
...
+
+    :return: Exception if caught, None otherwise
+    :rtype: Optional[Exception]
+    """
+
+    try:
+        logger.info(f"Clearing workspace root: {workspace_root}")
+        p4_connection.client = workspace_name
+        p4_connection.run("sync", "-f", workspace_root + "/...#0")
+    except Exception as e:
```
baxeaz, line 319: For the top case at least I think we're dealing with a P4Exception?  Might be better to break that case out rather than fold all exceptions into it.  

Also it sure looks dangerous depending on the underlying error string for interpreting what type of exception you got, but based on https://help.perforce.com/helix-core/apis/p4python/current/Content/P4Python/python.classes.html#P4.P4Exc maybe that's the best practice?

src/deadline/unreal_perforce_utils/app.py
```diff
@@ -0,0 +1,354 @@
...
+    Build and return the workspace name based on the given project name:
+    <USERNAME>_<HOST>_<PROJECT_NAME>
+
+    :param project_name: Name of the project
+
+    :return: Workspace name
+    :rtype: str
+    """
+
+    return f"{os.getlogin()}_{socket.gethostname()}_{project_name}"
```
baxeaz: I don't think this will work well in EC2 with multiple workers - I think hostname collisions are likely to happen fairly often (Creating additional workers from an AMI for example).  I think we need to include instance id, or fleet/worker id, or some unique or something to make sure we don't run into collision

src/deadline/unreal_perforce_utils/app.py
```diff
@@ -0,0 +1,354 @@
...
+            return e
+
+    return None
+
+
+def clear_workspace_files(
+    workspace_name: str, workspace_root: str, p4_connection: perforce.P4
+) -> Optional[Exception]:
+    """
+    Delete all local files from workspace syncing them to revision #0.
```
baxeaz, line 301: If we're about to delete this workspace, is this necessary?  Is there a server side issue if we don't do this or is this just attempting to clean up our client filesystem efficiently?  We definitely don't want to just leave a bunch of files hanging around - I made a comment offline about whether this is intended to be temporary currently, and if so if we could just put it under the session directory instead so cleanup happened there (Along with any other files that got created in the workspace which wouldn't be handled by this currently).

src/deadline/unreal_perforce_utils/app.py
```diff
@@ -0,0 +1,354 @@
...
+            "since no workspace name or Unreal project name is provided"
+        )
+
+    p4 = perforce.PerforceConnection().p4
+
+    last_exception = None
+
+    workspace_root = p4.fetch_client(workspace_name_to_delete).get("Root").replace("\\", "/")
+    if workspace_root and os.path.exists(workspace_root):
+        revert_exc = revert_all_changes_in_workspace(
```
baxeaz, line 368: Is this because our workspace deletion below could fail if we've got files checked out?  (Just want to know for the future - might be worth a comment about why this and clear workspace below are important).

src/deadline/unreal_perforce_utils/cli.py
```diff
@@ -0,0 +1,51 @@
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
baxeaz: Just a test/helper script?

src/deadline/unreal_perforce_utils/perforce.py
```diff
@@ -0,0 +1,218 @@
...
+            p4.connect()
+        except P4Exception as e:
+            raise exceptions.PerforceConnectionError(
+                f"Could not connect Perforce server {p4.port} as user {p4.user}\n{str(e)}"
+            )
+
+        p4.input = "y"
+        p4.run("trust", ["-y"])
+
+        p4_password = password or os.getenv("P4PASSWD")
```
baxeaz, line 58: Mentioned offline - do we have a current recommended way for this info (And P4PORT/P4USER) to get to the worker?  Idea was could P4PORT and P4USER optionally be sent as part of the job with an arn to a secrets manager secret for P4PASSWD

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
```diff
@@ -496,17 +500,22 @@ def __init__(
...
             self._name = self.mrq_job.job_name
 
-        ugs_envs = [
-            env for env in self._environments if isinstance(env, UnrealOpenJobUgsEnvironment)
-        ]
+        ugs_envs = [e for e in self._environments if isinstance(e, UgsUnrealOpenJobEnvironment)]
+        p4_envs = [e for e in self._environments if isinstance(e, P4UnrealOpenJobEnvironment)]
+        if ugs_envs and p4_envs:
+            raise exceptions.FailedToDetectFilesTransferStrategy(
+                "Failed to detect how to transfer project files to render because "
```
baxeaz, line 512: I don't think we should go back and change this as part of this PR, but now that all of the pieces are together it does feel a bit odd that these two environments don't (optionally) work together.  Like at a high level it seems like Ugs should depend on the perforce environment, and perforce without the ugs environment just works as native perforce.  Maybe there were technical reasons why that didn't make sense though?

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
```diff
@@ -745,34 +750,106 @@ def _update_environments_settings_from_mrq_job(
...
+            file_ext=".json",
+        )
+        parameter_values = RenderUnrealOpenJob.update_job_parameter_values(
+            job_parameter_values=parameter_values,
+            job_parameter_name=OpenJobParameterNames.PERFORCE_WORKSPACE_SPECIFICATION_TEMPLATE,
+            job_parameter_value=workspace_spec_template,
+        )
+        self._asset_references.input_filenames.add(workspace_spec_template)
+
+        # We need to collect job dependencies on the Artist node because some of them of
```
baxeaz, line 881: Ok I think I asked about this one offline as well - what sorts of files fall into this category again?  Is this essentially "Everything else" that we don't sync initially if I'm using p4 without UGS?

baxeaz:
Going to give it another pass, but wanted to send everything I had so far!

docs/source/docs_usage/use_cases/create_p4_render_job.rst
Iskander7d: Yep, my IDE just ignored this files for some reason. Added them in latest commit

alf-devnull: Yep, my IDE just ignored these files for some reason. Added them in latest commit

docs/source/docs_usage/use_cases/create_p4_render_job.rst
alf-devnull: fixed

docs/source/docs_usage/use_cases/create_p4_render_job.rst
alf-devnull: fixed

docs/source/docs_usage/use_cases/create_p4_render_job.rst
alf-devnull: Fixed.
Yes. All of the jobs have Launch environment

src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
alf-devnull: Thats right. Actually, if Unreal subprocess no longer exist or it has the .poll() != None, we should check the return code before treating it as an error. For example, If artists will run their custom script (with "Script" handler) that quit the editor normally in the end, we should not treat it as an error.

So I think its better to change this comment. Any thoughts?

src/deadline/unreal_adaptor/UnrealClient/unreal_client.py
alf-devnull: UnrealClient inherits from WinClientInterface which in turn inherits from BaseClientInterface (https://github.com/OpenJobDescription/openjd-adaptor-runtime-for-python/blob/mainline/src/openjd/adaptor_runtime_client/base_client_interface.py).
And it has this log in the poll() method.

We just added an extra check to avoid having to process the request on every tick of the slate and didn't change the rest of the code to match the parent interface.

src/deadline/unreal_adaptor/UnrealClient/unreal_client.py
alf-devnull: fixed

src/deadline/unreal_perforce_utils/app.py
alf-devnull: Yes, thats right. Thanlks for noticing. Will think how to refactor it.

Maybe at least add the ability to create a workspace from a dictionary and a json string, in case we want to reuse this method somewhere else where there is no need to convert to a string and replace occurrences. In general, I'll think about how best to do this.

In fact, the template is often simple and consists of the workspace name and stream name. Less often, instead of a stream, there can be a list of views, where there can be a lot of occurrences of {workspace_name}, so we do this "dirty" but quick convertation to a string to quickly replace everything.

src/deadline/unreal_perforce_utils/app.py
alf-devnull: I know it looks bad, but P4 API raise the error if file is up to date that not a real error (for example, you try to sync some file which is up to date). Furthermore, as I know, P4 API does not have other exception classes that we can rely on. In all the time I've been working with P4, only one exception has been thrown out - P4Exception.

And we just make safe check if we trying to sync the workspace to 0 revision (delete local files) which already has no files or some files are missed

src/deadline/unreal_perforce_utils/app.py
alf-devnull: What the OS environment variables for Instance ID/Worker ID?

alf-devnull: Sure, will add Worker ID to it

src/deadline/unreal_perforce_utils/app.py
alf-devnull: Will add the example of template

src/deadline/unreal_perforce_utils/app.py
alf-devnull: P4 delete workspace operation just tell the P4 server that this folder on this machine is no longer attached to P4 workspace and delete the workspace from the server. It leave all of the files here.

And its much better and safer to delete files with P4 API (sync to revision 0) than just clear the system files to prevent any server-side issues during the next sync, yes thats right. 

Unreal may produce some files in temporary files which are not synced to the P4 depot, so its okay to leave them here. Furthermore, most of them would be regenerated on the next launch.

As we discussed earlier, I think we will add the check if P4_CLIENTS_ROOT_DIRECTORY os environ variable is set. And if so, skip the files deletion since we want to leave them for the next jobs and decrease the sync time

src/deadline/unreal_perforce_utils/app.py
alf-devnull: No, actually, we run delete with -f (force) option, so P4 will not produce the errors about checked out files. But to be more defensive, we also revert all local changes before deletion.

In other words, we roll back all possible changes in reverse order:

On start: Create workspace -> Sync files -> Produce local changes (optionally)
On end: Delete worksapce <- Delete files <- Revert changes

src/deadline/unreal_perforce_utils/cli.py
alf-devnull: No, actually this cli.py used for creating the executable `unreal-engine-p4-utils.exe` which is used in P4 Sync OpenJD Environment

src/deadline/unreal_perforce_utils/perforce.py
alf-devnull: The best way is to provide P4PORT and P4USER. 
P4CLIENT will be created automatically
P4PASSWD is optional, across a lot of render farms which use P4 there is single service user with infinite session on the render nodes.

But yes, its a good idea to send the P4PASSWD via secret manager if you want to create new connection each time in case of SMF, for example

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
alf-devnull: Technical reasons come from:
1. the way of syncing the files
2. The main idea of universal proposal on customizing the job with steps

From the beginning, we are trying to avoid the duplication of the entities in CPP and Python.

In theory, the submitter should not depend on what jobs, steps and environments are submitted. At the moment, we have 5 basic entities (job, render job, step, render step, environment).
We had to add render job and render step, because they have completely different logic. Accordingly, to determine that this is a render, we create the same entities in the CPP.
If we want to know exactly what strategy we are using, then we need to add specific implementations for file transfer (P4 or UGS sync), then they will need to be duplicated. This results in a lot of unnecessary entities that will not have a special implementation inside.

UGS and Perforce imply completely different operations and pipelines for the same job. The first synchronizes the entire workspace, downloads engine binaries, and uses another utility for this in C#. The second synchronizes only project files, and uses the Python API directly. They have different approaches, so there is no point in building their hierarchy.

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
alf-devnull: So the files for pure P4 is:
1. Workspace specification template for creating the same workspace as the artist's one. UGS do it out of the box and have more strict rules for creating the workspaces, their naming, etc.
2. Job dependencies descriptor (assets to sync). Because we can have "soft" dependencies, which cannot be collected on the worker node because their referencer dont know about the path if they are not exist on the disk.
For default Render, UGS and P4:
1. Extra CMD Args file
2. Serialized MRQ file

src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
baxeaz: Ah yeah makes sense!  Yeah updating the comment with something like that description you gave should work

src/deadline/unreal_perforce_utils/app.py
baxeaz: Makes sense, thanks

src/deadline/unreal_perforce_utils/perforce.py
baxeaz: Got it - can we add this info to the setup instructions?

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
baxeaz: Ok thanks for the explanation!

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
baxeaz: Got it, thanks - can we add those to the comment?

alf-devnull: Added

src/deadline/unreal_perforce_utils/app.py
alf-devnull: Added

src/deadline/unreal_perforce_utils/app.py
alf-devnull: Comments updated

src/deadline/unreal_perforce_utils/app.py
alf-devnull: Dumps, Loads removed. Go through possible "View"s and replace instead

src/unreal_plugin/Content/Python/openjd_templates/render_job.yml
```diff
@@ -5,13 +5,18 @@ parameterDefinitions:
   type: PATH
   objectType: FILE
   dataFlow: IN
+  userInterface:
+    control: HIDDEN
 - name: ExtraCmdArgs
   type: STRING
   default: '-log'
+<<<<<<<< HEAD:src/unreal_plugin/Content/Python/openjd_templates/job_template.yml
```
KaboomFox: Merge issue

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_custom_step_handler.py
```diff
@@ -60,18 +60,34 @@ def validate_script(script_path: str) -> ModuleType:
...
+
+                # If the command ran successfully, this will return None else the
+                # failure
+                # https://dev.epicgames.com/documentation/en-us/unreal-engine/python-api/class/PythonScriptLibrary?application_version=5.4#unreal.PythonScriptLibrary.execute_python_command_ex
+                if failure:
+                    raise RuntimeError(failure)
+
+            logger.info(f"Custom Step Executor Result: {result}")
             return True
+
```
karthikbekalp, line 90: Nit: It's generally better to have exception handling for specific exceptions that can be raised, rather than using a broad Exception catch-all. This approach allows for better error handling and more precise error messaging. However, this is not a blocker for the current PR.

src/deadline/unreal_perforce_utils/app.py
```diff
@@ -0,0 +1,387 @@
...
+
+    if not os.path.exists(workspace_specification_template_path):
+        raise FileNotFoundError(
+            f"The workspace specification template does not exist: {workspace_specification_template_path}"
+        )
+
+    logger.info(
+        f"Getting workspace specification template from file: {workspace_specification_template_path} ..."
+    )
+    with open(workspace_specification_template_path, "r") as f:
```
karthikbekalp: We have observed that Windows systems read files using Windows-specific encoding rather than UTF-8 when using the open() function. Can we modify this to use with `open(workspace_specification_template_path, "r", encoding="utf8") as f`: to properly handle non-ASCII characters in the job template?

karthikbekalp (commented):
Few comments but nothing blocking. 


Although I see that there are few tests that have been failing. Once we fix that, we should be good to ship this. :)

src/deadline/unreal_perforce_utils/app.py
alf-devnull: added

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_custom_step_handler.py
alf-devnull: handle keyerror and runtime error

karthikbekalp: Thank you for adding the exception handling. However, we still need the broad exception handler to catch exceptions beyond KeyError and RuntimeError. 

Additionally, could we improve the error messages to include the specific reason for KeyError? This would help customers troubleshoot configuration issues more effectively. 

Similarly, a better messaging would be useful for RuntimeError cases as well.

src/deadline/unreal_perforce_utils/app.py
karthikbekalp: Thanks!

baxeaz (approved):
Have submitted successful renders locally both with JA and Perforce.  Thanks for all the hard work on this!

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_custom_step_handler.py
karthikbekalp: Not blocking the PR on this.

karthikbekalp (approved):

PR #121, opened 2025-02-19
test: Adding an integration test which submits a level sequence job

baxeaz:
We need more thorough integration testing of creating a job successfully from within the editor and submitting it through the submitter.

Add a CreateJob integration test

Easier to validate changes

Ran the test successfully from within the UI and on the command line.  Tested multiple submissions.

Yes

No

No

----

### DIFF
--- a/src/unreal_plugin/Source/UnrealDeadlineCloudService/Private/Tests/Integration/DeadlinePluginTest_CreateJobTest.cpp
+++ b/src/unreal_plugin/Source/UnrealDeadlineCloudService/Private/Tests/Integration/DeadlinePluginTest_CreateJobTest.cpp
@@ -1,0 +1,201 @@
+#include "CoreMinimal.h"
+#include "HAL/PlatformTime.h"
+#include "Misc/AutomationTest.h"
+#include "Tests/AutomationCommon.h"
+#include "LevelSequence.h"
+#include "MoviePipelineQueueSubsystem.h"
+#include "MoviePipelineQueue.h"
+#include "MovieScene.h"
+#include "MovieRenderPipelineSettings.h"
+#include "MoviePipelineEditorBlueprintLibrary.h"
+#include "MovieRenderPipeline/DeadlineCloudRenderStepSetting.h"
+#include "MovieRenderPipeline/MoviePipelineDeadlineCloudExecutorJob.h"
+#include "Modules/ModuleManager.h"
+
+DEFINE_LOG_CATEGORY_STATIC(LogCreateJobTest, Log, All);
+
+// Path to the level sequence Asset to attempt to create a job for
+const char LevelSequencePath[] = "/Game/Levels/Main_SEQ.Main_SEQ";
+
+class WaitForJobCreationLogCommand : public IAutomationLatentCommand, public FOutputDevice
+{
+    // Test command for registering/deregistering log listeners, running a render job using the provided queue and executor, and
+    // listening for expected logging output to indicate success
+public:
+    WaitForJobCreationLogCommand(FAutomationTestBase* testInstance, UMoviePipelineQueueSubsystem* queueSubsystem, UMoviePipelineExecutorBase* executorBase)
+        : m_startTime(FPlatformTime::Seconds())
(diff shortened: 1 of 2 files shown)


src/unreal_plugin/Source/UnrealDeadlineCloudService/Private/Tests/Integration/DeadlinePluginTest_CreateJobTest.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+    UE_LOG(LogCreateJobTest, Display, TEXT("Creating executor"));
+    UMoviePipelineExecutorBase* executorBase = NewObject<UMoviePipelineExecutorBase>(GetTransientPackage(), ExecutorClass);
+
+    // Command to set up our log listeners and run our job
+    ADD_LATENT_AUTOMATION_COMMAND(WaitForJobCreationLogCommand(testInstance, QueueSubsystem, executorBase));
+
+    // Cleanup command to restore our queue to its original state
+    ADD_LATENT_AUTOMATION_COMMAND(RestoreQueueCommand(QueueSubsystem, OriginalQueue));
+
```
crowecawcaw, line 198: Is there anything to assert in here? Are mostly just interested if the job can succeed?

crowecawcaw (approved):
How is this test triggered? Would it make sense to add instructions?

src/unreal_plugin/Source/UnrealDeadlineCloudService/Private/Tests/Integration/DeadlinePluginTest_CreateJobTest.cpp
```diff
@@ -0,0 +1,201 @@
...
+#include "MovieRenderPipelineSettings.h"
+#include "MoviePipelineEditorBlueprintLibrary.h"
+#include "MovieRenderPipeline/DeadlineCloudRenderStepSetting.h"
+#include "MovieRenderPipeline/MoviePipelineDeadlineCloudExecutorJob.h"
+#include "Modules/ModuleManager.h"
+
+DEFINE_LOG_CATEGORY_STATIC(LogCreateJobTest, Log, All);
+
+// Path to the level sequence Asset to attempt to create a job for
+const char LevelSequencePath[] = "/Game/Levels/Main_SEQ.Main_SEQ";
```
joel-wong-aws, line 18: What is this file? Does it need to be committed to the repo?

src/unreal_plugin/Source/UnrealDeadlineCloudService/Private/Tests/Integration/DeadlinePluginTest_CreateJobTest.cpp
baxeaz: The test can be found with the other Unreal Automation tests following https://github.com/aws-deadline/deadline-cloud-for-unreal-engine/blob/mainline/DEVELOPMENT.md#running-unreal-spec-tests  (This one is under a new "Integration" subheading)

I'm working on adding additional support right now which will also give a simplified way of running it all from pytest (Along with building/installing the code, validating that you've got the worker agent running, watching the job output, etc).

As far as assertions that "WaitForJobCreationLogCommand" is what's doing that - Unreal sends the request off to the python layer (the submitter) and we're registering and watching for output to say we seem to have created a job and would have reported it to the user in a dialog.

src/unreal_plugin/Source/UnrealDeadlineCloudService/Private/Tests/Integration/DeadlinePluginTest_CreateJobTest.cpp
baxeaz: This is temporary until we have assets which we can commit to the repo.  I've got a request in to Fuzzy Pixel for just those assets!  Until then this test can be run by changing this to point at a level sequence from any given project/scene you'd like - this one happens to correspond to Meerkat which is the most common test level folks use currently.

joel-wong-aws (approved):

Issue #126, opened 2025-02-28
Bug: build_plugin.py: error: unrecognized arguments: --install --worker


komapa:
### Describe the bug

```
PS C:\Users\kangov> git clone https://github.com/aws-deadline/deadline-cloud-for-unreal-engine.git
Cloning into 'deadline-cloud-for-unreal-engine'...
remote: Enumerating objects: 1451, done.
remote: Counting objects: 100% (435/435), done.
remote: Compressing objects: 100% (297/297), done.
remote: Total 1451 (delta 206), reused 185 (delta 136), pack-reused 1016 (from 3)
Receiving objects: 100% (1451/1451), 574.12 KiB | 11.48 MiB/s, done.
Resolving deltas: 100% (347/347), done.
Filtering content: 100% (57/57), 2.08 MiB | 1.92 MiB/s, done.
PS C:\Users\kangov> cd deadline-cloud-for-unreal-engine
PS C:\Users\kangov\deadline-cloud-for-unreal-engine> git switch release
Filtering content: 100% (38/38), 2.81 MiB | 2.73 MiB/s, done.
branch 'release' set up to track 'origin/release'.
Switched to a new branch 'release'
PS C:\Users\kangov\deadline-cloud-for-unreal-engine> python scripts/build_plugin.py --install --worker
usage: build_plugin.py [-h] [--runuat-path RUNUAT_PATH] [--input-folder INPUT_FOLDER] [--output-folder OUTPUT_FOLDER]
                       [--upload-bucket UPLOAD_BUCKET]
build_plugin.py: error: unrecognized arguments: --install --worker
```

### Expected Behaviour

To not fail :)

### Current Behaviour

Because the current release does not have the newest code, the documentation from `master` branch is incorrect in terms of asking to select the release branch.

### Reproduction Steps

```
PS C:\Users\kangov> git clone https://github.com/aws-deadline/deadline-cloud-for-unreal-engine.git
Cloning into 'deadline-cloud-for-unreal-engine'...
remote: Enumerating objects: 1451, done.
remote: Counting objects: 100% (435/435), done.
remote: Compressing objects: 100% (297/297), done.
remote: Total 1451 (delta 206), reused 185 (delta 136), pack-reused 1016 (from 3)
Receiving objects: 100% (1451/1451), 574.12 KiB | 11.48 MiB/s, done.
Resolving deltas: 100% (347/347), done.
[...]

PR #127, opened 2025-02-28
docs: Adding note about choosing branch and corresponding instruction…

baxeaz:
Midway through our setup instructions users are asked to choose between the release and mainline branch.  In many cases they may be reading the mainline version and switch to the release branch where some of the following steps, or even steps they've already completed are no longer correct.

This has been raised in #126  

Add a note at the top of the instructions about branch choice

Less frustration and time wasted following the incorrect steps

Docs change

Docs only

Docs only

No

----


karthikbekalp (approved):
Cool, thanks for fixing this!

crowecawcaw (approved):

PR #130, opened 2025-03-10
feat: support AWS Secrets Manager

alf-devnull:
When using Perforce /UGS, it is necessary to pass sensitive parameters such as password, user, port. It is necessary to minimize their transfer through Job environment variables and use a special secret manager to increase the security of user data

- Provide "AWS_SECRET_P4INFO" environment variable that contains name of the secret that need to find to retrieve P4 connection parameters (any of them).
- Use this environment variable on the worker side to retrieve connection parameters using Boto3 client for SecretsManager service.

Now all of the users' sensitive data, such as P4 Password can be applied without passing within Job environment

- unit tests
- smoke tests
- Have you run the unit tests? Yes

Yes

Yes

No
----

### DIFF
--- a/docs/source/docs_usage/p4_credentials_management.rst
+++ b/docs/source/docs_usage/p4_credentials_management.rst
@@ -30,6 +30,73 @@ the new connection will be configured as:
    User = user.from.registry
    Password = password.from.connection
 
+Use AWS Secrets Manager
+***********************
+
+AWS Secrets Manager allows you to store and manage secrets, such as connection credentials,
+in a secure, highly available, and easily managed service. All workers in your fleet (CMF or SMF)
+can access it if you have configured roles and access across workers in your farm.
+
+To make worker apply connection credentials from AWS Secrets Manager, you need to provide the
+name of the secret in the ``AWS_SECRET_P4INFO`` environment variable in any P4 (UGS) Sync Environment you use.
+Or create the new Environment template and prepend it to your Job.
+
+.. code-block:: yaml
+
+   name: P4SyncCmf
+   variables:
+     AWS_SECRET_P4INFO: DeadlineCloud/MyFleet/P4INFO
+
+Secret ``DeadlineCloud/MyFleet/P4INFO`` may contains any of the following key/value pairs:
+
+#. P4PORT - Perforce server port
+#. P4USER - Perforce user
+#. P4PASSWD - Perforce password
+
+.. note:: The names of the keys should be exactly the same
+   as the P4 connection parameters, i. e. ``P4PORT``, ``P4USER``, ``P4PASSWD``
+
+Example:
+
+You have next P4 Sync Environment:
+
+.. code-block:: yaml
+
+   name: P4SyncCmf
+   variables:
+     AWS_SECRET_P4INFO: DeadlineCloud/MyFleet/P4INFO
+     P4USER: j.doe
+     P4PORT: ssl:my-perforce.com:1666
+   ...
+
+And the secret ``DeadlineCloud/MyFleet/P4INFO`` contains:
+
+#. P4PASSWD = MyVeRyS3cretP4ssW0rd
+#. P4USER = aws-admin-user
+
+Environment variables from AWS Secrets Manager will override the ones from the Job. Configured connection
+parameters will be:
+
+.. code-block::
+
+   Port = ssl:my-perforce.com:1666
+   User = aws-admin-user
+   Password = MyVeRyS3cretP4ssW0rd
+
+User appears twice - in Job environment and in AWS Secrets Manager, but last one take precedence.
+
+.. warning::
+   In UGS case you need to use environment data asset for applying P4 secrets -
+   ``src/unreal_plugin/Content/OpenJD_DataAssets/Render/OpenJD_Environment_ApplyP4Secrets.uasset`` -
+   which is referencing to ``src/unreal_plugin/Content/Python/openjd_templates/p4/p4_apply_secrets_environment.yml``.
+
+   This environment apply P4 credentials from AWS Secretes Manager by printing them to the Job log
+   with prefix ``openjd_env:``. Please, see `OpenJD Environment`_ documentation about sharing new
+   environment variables across actions and other environments in runtime. AWS Deadline Cloud
+   constantly evolving and will soon receive an update that will eliminate the need to print variables to set them.
+   But for now it is the only option.
+
+.. _OpenJD Environment: https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/wiki/2023-09-Template-Schemas.md#4-environment
 
 Pass Connection Credentials within the Job Environment
 ******************************************************
@@ -107,6 +174,8 @@ the `Deadline Cloud Queue Environments Git sample`_:
 Conclusions
 ***********
 
+#. The best option is to use AWS Secrets Manager to store connection credentials. It is the most secure,
+   and you can use it for both CMF and SMF.
 #. If you have a CMF farm, where all workers are inside a network that is not accessible from the outside,
    then you can use a single P4 admin user on all workers, changing only ``P4PORT`` if necessary.
    Optionally, you can pass the credentials within the Job Environment.
--- a/docs/source/docs_usage/p4_credentials_management.rst
+++ b/docs/source/docs_usage/p4_credentials_management.rst
@@ -37,6 +37,27 @@ AWS Secrets Manager allows you to store and manage secrets, such as connection c
 in a secure, highly available, and easily managed service. All workers in your fleet (CMF or SMF)
 can access it if you have configured roles and access across workers in your farm.
 
+.. note:: You should configure the access for action ``secretsmanager:GetSecretValue`` for the worker in the queue.
+   For more information visit `Manage access to Windows job user secrets`_. In general, a request to configure access looks like this:
+
+   .. code-block:: json
+
+      {
+        "Version" : "2012-10-17",
+        "Statement" : [
+          //...
+          {
+            "Effect" : "Allow",
+            "Action" : "secretsmanager:GetSecretValue",
+            "Resource" : [
+              "arn:aws:secretsmanager:<your_region>:<your_resource_account>:secret:<your_secret_id_or_wildcard>"
+          }
+          //...
+        ]
+      }
+
+.. _Manage access to Windows job user secrets: https://docs.aws.amazon.com/deadline-cloud/latest/developerguide/manage-access-windows-secrets.html#grant-access
+
 To make worker apply connection credentials from AWS Secrets Manager, you need to provide the
 name of the secret in the ``AWS_SECRET_P4INFO`` environment variable in any P4 (UGS) Sync Environment you use.
 Or create the new Environment template and prepend it to your Job.
@@ -47,6 +68,8 @@ Or create the new Environment template and prepend it to your Job.
    variables:
      AWS_SECRET_P4INFO: DeadlineCloud/MyFleet/P4INFO
 
+.. note:: If ``AWS_SECRET_P4INFO`` value is empty, applying connections from Secrets Manager will be skipped
+
 Secret ``DeadlineCloud/MyFleet/P4INFO`` may contains any of the following key/value pairs:
 
 #. P4PORT - Perforce server port
@@ -56,9 +79,7 @@ Secret ``DeadlineCloud/MyFleet/P4INFO`` may contains any of the following key/va
 .. note:: The names of the keys should be exactly the same
    as the P4 connection parameters, i. e. ``P4PORT``, ``P4USER``, ``P4PASSWD``
 
-Example:
-
-You have next P4 Sync Environment:
+For example, you have next P4 Sync Environment:
 
 .. code-block:: yaml
 
@@ -85,18 +106,25 @@ parameters will be:
 
 User appears twice - in Job environment and in AWS Secrets Manager, but last one take precedence.
 
-.. warning::
-   In UGS case you need to use environment data asset for applying P4 secrets -
-   ``src/unreal_plugin/Content/OpenJD_DataAssets/Render/OpenJD_Environment_ApplyP4Secrets.uasset`` -
-   which is referencing to ``src/unreal_plugin/Content/Python/openjd_templates/p4/p4_apply_secrets_environment.yml``.
+In UGS case you need to prepend to the Job's environments data asset for applying P4 secrets -
+``src/unreal_plugin/Content/OpenJD_DataAssets/Render/OpenJD_Environment_ApplyP4Secrets.uasset`` -
+which is referencing to ``src/unreal_plugin/Content/Python/openjd_templates/p4/p4_apply_secrets_environment.yml``:
 
-   This environment apply P4 credentials from AWS Secretes Manager by printing them to the Job log
+.. literalinclude:: ../../../src/unreal_plugin/Content/Python/openjd_templates/p4/p4_apply_secrets_environment.yml
+    :language: yaml
+    :linenos:
+
+.. warning:: This environment apply P4 credentials from AWS Secretes Manager by printing them to the Job log
    with prefix ``openjd_env:``. Please, see `OpenJD Environment`_ documentation about sharing new
-   environment variables across actions and other environments in runtime. AWS Deadline Cloud
-   constantly evolving and will soon receive an update that will eliminate the need to print variables to set them.
-   But for now it is the only option.
+   environment variables across actions and other environments in runtime.
+
+   **Use with caution because sensitive data, such as the password, port, and user,**
+   **will be reflected in the job execution history.**
 
 .. _OpenJD Environment: https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/wiki/2023-09-Template-Schemas.md#4-environment
+
+You can override ``AWS_SECRET_P4INFO`` value in MRQ Slate UI (Environments overrides block) or update its value
+in the ``src/unreal_plugin/Content/Python/openjd_templates/p4/p4_apply_secrets_environment.yml`` file.
 
 Pass Connection Credentials within the Job Environment
 ******************************************************
--- a/docs/source/docs_usage/p4_credentials_management.rst
+++ b/docs/source/docs_usage/p4_credentials_management.rst
@@ -118,10 +118,18 @@ which is referencing to ``src/unreal_plugin/Content/Python/openjd_templates/p4/p
    with prefix ``openjd_env:``. Please, see `OpenJD Environment`_ documentation about sharing new
    environment variables across actions and other environments in runtime.
 
-   **Use with caution because sensitive data, such as the password, port, and user,**
-   **will be reflected in the job execution history.**
+   **Consider adding a CloudWatch data protection policy to your account if you'll be echoing***
+   **sensitive information. For example, this will apply a global policy to your CloudWatch logs**
+   **which suppresses "openjd_env": lines which appear to be setting environment variables:**
+
+   .. literalinclude:: ../resources/logs_policy_example.sh
+    :language: sh
+    :linenos:
+
+   For more information and options please see `Custom data identifiers`_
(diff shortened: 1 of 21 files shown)


docs/source/docs_usage/p4_credentials_management.rst
```diff
@@ -30,6 +30,73 @@ the new connection will be configured as:
...
+
+.. warning::
+   In UGS case you need to use environment data asset for applying P4 secrets -
+   ``src/unreal_plugin/Content/OpenJD_DataAssets/Render/OpenJD_Environment_ApplyP4Secrets.uasset`` -
+   which is referencing to ``src/unreal_plugin/Content/Python/openjd_templates/p4/p4_apply_secrets_environment.yml``.
+
+   This environment apply P4 credentials from AWS Secretes Manager by printing them to the Job log
+   with prefix ``openjd_env:``. Please, see `OpenJD Environment`_ documentation about sharing new
+   environment variables across actions and other environments in runtime. AWS Deadline Cloud
+   constantly evolving and will soon receive an update that will eliminate the need to print variables to set them.
```
baxeaz: We can't include this unfortunately - can you remove this and the last sentence (But for now it is the only option)?

alf-devnull: We cant include description of printing mechanism, right? If so, should we warn about it anyway but with other words?
And should we also remove warnings from _src/unreal_plugin/Content/Python/openjd_templates/p4/p4_apply_secrets_environment.yml_ and `deadline.unreal_perforce_utils.app.apply_perforce_secrets`

baxeaz: The warnings are ok, customers absolutely should know how it currently works!  It's just the "AWS Deadline Cloud
   constantly evolving and will soon receive an update that will eliminate the need to print variables to set them.  But for now it is the only option" part that we can't include as we can't predict the timing on that currently

docs/source/docs_usage/p4_credentials_management.rst
```diff
@@ -107,6 +174,8 @@ the `Deadline Cloud Queue Environments Git sample`_:
 Conclusions
 ***********
 
+#. The best option is to use AWS Secrets Manager to store connection credentials. It is the most secure,
```
baxeaz: Can we phrase this as - "Perforce connection credentials can be stored in centralized location in Secrets Manager from a secret your workers will retrieve at runtime.  Be aware that with UGS this method currently requires exposing these credentials to your job logs in your account.  Consider using a data protection policy to hide them."? (And link to our guidance)  "Best" is tough to say, since some customers will already have everything already configured on their CMF workers

docs/source/docs_usage/p4_credentials_management.rst
```diff
@@ -30,6 +30,101 @@ the new connection will be configured as:
...
+.. literalinclude:: ../../../src/unreal_plugin/Content/Python/openjd_templates/p4/p4_apply_secrets_environment.yml
+    :language: yaml
+    :linenos:
+
+.. warning:: This environment apply P4 credentials from AWS Secretes Manager by printing them to the Job log
+   with prefix ``openjd_env:``. Please, see `OpenJD Environment`_ documentation about sharing new
+   environment variables across actions and other environments in runtime.
+
+   **Use with caution because sensitive data, such as the password, port, and user,**
+   **will be reflected in the job execution history.**
```
baxeaz: Can we add a suggestion here of something like "Consider adding a CloudWatch data protection policy to your account if you'll be echoing sensitive information.  For example, this will apply a global policy to your CloudWatch logs which suppresses openjd_env: lines which appear to be setting environment variables.

aws logs put-account-policy \
  --policy-name "MaskOpenJDEnvSecrets" \
  --policy-type DATA_PROTECTION_POLICY \
  --scope ALL \
  --policy-document '{  
    "Name": "openjd_data_protection",
    "Description": "Hides openjd_env log line data which may be setting sensitive info",
    "Version": "2021-06-01",
    "Configuration": {
      "CustomDataIdentifier": [
        {"Name": "OpenjdEnv", "Regex": "openjd_env:\\s*[^=]+=.*"}   
      ]
    },
    "Statement": [
		{
            "Sid": "audit-policy",
            "DataIdentifier": [
                "OpenjdEnv"
            ],
            "Operation": {
                "Audit": {
                    "NoFindingsDestination": {
                    }
                }
            }
        },
        {
            "Sid": "redact-policy",
            "DataIdentifier": [
            "OpenjdEnv"
            ],
            "Operation": {
                "Deidentify": {
                    "MaskConfig": {
                    }
                }
            }
        }
    ]
}'

For more information and options please see https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/CWL-custom-data-identifiers.html

src/deadline/unreal_perforce_utils/app.py
```diff
@@ -386,3 +386,37 @@ def delete_workspace(workspace_name: Optional[str] = None, project_name: Optiona
...
+    by name stored in AWS_SECRET_P4INFO and apply all key/value pairs from it as environment variables.
+
+    The following environment variables can be set:
+
+    - P4USER
+    - P4PASSWD
+    - P4PORT
+
+    .. warning::
+       Use with caution! In current implementation all retrieved env variables will be applied with
```
baxeaz: And then here and the logger warning below I think we can point customers to that guidance as well, so something like:

Be aware that by default environment variables are persisted by printing to stdout (See the openjd_env: line at the end of this method).  Consider adding a CloudWatch data protection policy to prevent potentially sensitive information from being echoed to your logs.  (And link to our guidance)

src/deadline/unreal_perforce_utils/app.py
```diff
@@ -386,3 +386,37 @@ def delete_workspace(workspace_name: Optional[str] = None, project_name: Optiona
...
+       printing to stdout, some sensitive data can be reflected in the job execution history.
+
+    """
+
+    logger.warning(
+        "When using environment variables to connect to Perforce, they will appear in the logs "
+        "with the prefix 'openjd_env'. Use with caution because sensitive data, such as the "
+        "password, port, and user, will be reflected in the job execution history."
```
baxeaz: Once more sentence like above, "If you have not already done so, consider adding a CloudWatch data protection.."

src/deadline/unreal_perforce_utils/secret_manager.py
```diff
@@ -0,0 +1,91 @@
...
+
+    secret_env_name = "AWS_SECRET_P4INFO"
+    if os.getenv(secret_env_name) in [None, ""]:
+        logger.warning(
+            f"{secret_env_name} environment variable not found or empty. Cant get perforce info"
+        )
+        return None
+
+    p4_info_secret = get_secret_from_env(secret_env_name)
+    return ast.literal_eval(p4_info_secret) if p4_info_secret else None
```
baxeaz: Hmm this may throw an exception which I think in some cases might contain the info - I know we said to expect JSON which it should be, we should probably add some sort or protection though to be sure we don't unexpectedly echo anything.  Would also be good to have a test that validates that we aren't returning anything we don't expect either way

baxeaz (changes requested):

docs/source/docs_usage/p4_credentials_management.rst
alf-devnull: sure, will update

src/deadline/unreal_perforce_utils/secret_manager.py
alf-devnull: good take, will add

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_custom_step_handler.py
```diff
@@ -87,13 +87,21 @@ def run_script(self, args: dict) -> bool:
 
             logger.info(f"Custom Step Executor Result: {result}")
             return True
-
-        except (KeyError, RuntimeError) as e:
+        except KeyError as e:
```
karthikbekalp, line 90: Would we want a broad exception in case any other exception is raised?

src/deadline/unreal_perforce_utils/app.py
```diff
@@ -386,3 +386,37 @@ def delete_workspace(workspace_name: Optional[str] = None, project_name: Optiona
...
+    )
+    logger.info("Applying perforce secrets from Boto3 SecretsManager ...")
+
+    p4_info = secret_manager.get_perforce_info()
+    if not p4_info:
+        logger.info("No perforce secrets found in Boto3 SecretsManager. Skip applying")
+        return
+
+    for env_name, env_value in p4_info.items():
+        # For some reason, adaptor doesn't show logger records, need to R&D
```
karthikbekalp, line 424: Maybe add this as a TODO comment with a Github ticket? Just so that we have this somewhere written down?

src/deadline/unreal_perforce_utils/secret_manager.py
```diff
@@ -0,0 +1,91 @@
...
+    :rtype: Optional[str]
+    """
+    logger.info(f"Getting secret from environment variable: {secret_variable_name}")
+
+    sm_client = get_secret_manager_client()
+
+    secret_id = os.getenv(secret_variable_name)
+    if secret_id in [None, ""]:
+        logger.warning(f"Cant get secret from empty environment variable {secret_variable_name}")
+        return None
```
karthikbekalp: Should we raise an exception here?

src/deadline/unreal_perforce_utils/secret_manager.py
```diff
@@ -0,0 +1,91 @@
...
+    :type secret_variable_name: str
+    :return: The secret string
+    :rtype: Optional[str]
+    """
+    logger.info(f"Getting secret from environment variable: {secret_variable_name}")
+
+    sm_client = get_secret_manager_client()
+
+    secret_id = os.getenv(secret_variable_name)
+    if secret_id in [None, ""]:
```
karthikbekalp: nit: Hmm, wouldn't this be equivalent to just `if not secret_id`?

src/deadline/unreal_perforce_utils/secret_manager.py
```diff
@@ -0,0 +1,91 @@
...
+        return None
+
+    try:
+        response = sm_client.get_secret_value(SecretId=secret_id)
+    except ClientError as e:
+        logger.warning(
+            f"Failed to get secret from Boto3 SecretsManager by id {secret_id}."
+            f"{e} -- {traceback.format_exc()}"
+        )
+        return None
```
karthikbekalp: Same comment from above, should we raise an exception here?

src/deadline/unreal_perforce_utils/secret_manager.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+    Retrieves perforce connection parameters from Boto3 SecretsManager
+    using AWS_SECRET_P4INFO environment to get secret name.
+
+    :return: The perforce connection parameters
+    :rtype: Optional[dict[str, str]]
+    """
+
+    secret_env_name = "AWS_SECRET_P4INFO"
+    if os.getenv(secret_env_name) in [None, ""]:
```
karthikbekalp: nit: Similar nit as above.

src/deadline/unreal_perforce_utils/secret_manager.py
```diff
@@ -0,0 +1,91 @@
...
+        logger.warning(
+            f"{secret_env_name} environment variable not found or empty. Cant get perforce info"
+        )
+        return None
```
karthikbekalp, line 113: Should this raise an exception?

karthikbekalp (commented):
Added few minor comments. Thanks for your changes!

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_custom_step_handler.py
alf-devnull: sure, sorry, forgot to add `except Exception as e:`. will add

src/deadline/unreal_perforce_utils/app.py
alf-devnull: sure, will add

src/deadline/unreal_perforce_utils/secret_manager.py
alf-devnull: I think no, because users can dont use the secret manager to pass the connection credentials. In that case we just check if its None and make safe get on empty dict {} in places where we use that

src/deadline/unreal_perforce_utils/secret_manager.py
alf-devnull: yeah, Python have syntax sugar for checking that cases and representing None and empty strings and boolean False. But it that place and in other we just want be offensive

src/deadline/unreal_perforce_utils/secret_manager.py
alf-devnull: hmm. Maybe yes. But for that we need to check if AWS_SECRET_P4INFO in os environ on the top. And if it is in, we expect that we should retrieve something from SM. Good point, will update

src/deadline/unreal_perforce_utils/secret_manager.py
alf-devnull: And here we shouldn't, because the job can be without that variable. Conclusion: If that variable exists, raise exceptions if any in inner methods, return None otherwise

alf-devnull: General comment: When we trying to get p4 info, we expect that it can be an empty string. That means, users use default template where that variable is set to "". We should not raise an expection because that means SM feature just ignored

src/deadline/unreal_perforce_utils/secret_manager.py
karthikbekalp: Cool, makes sense. Thanks for clarifying.

src/deadline/unreal_perforce_utils/secret_manager.py
karthikbekalp: Cool, thanks for clarifying.

docs/source/docs_usage/p4_credentials_management.rst
alf-devnull: fixed

docs/source/docs_usage/p4_credentials_management.rst
alf-devnull: added

src/deadline/unreal_perforce_utils/app.py
alf-devnull: added

src/deadline/unreal_perforce_utils/app.py
alf-devnull: Just duplicate the warning from docstring. I thought it was unnecessary to repeat "consider adding a CloudWatch data protection"

src/deadline/unreal_perforce_utils/secret_manager.py
alf-devnull: added the validation, refactor a little the main approach - get_secret from secret name instead of get_secret_from_env from env name. So we don't need to double check the existance of the variables. And it allows us to raise exceptions in get method to prevent non valid data printing

src/deadline/unreal_perforce_utils/unreal_source_control.py
```diff
@@ -0,0 +1,179 @@
...
+    Validate that the given configparser object has the specified section.
+
+    :param sc_settings: A configparser object containing the source control settings.
+    :param section_name: The name of the section to validate.
+    :return: The configparser object if the section is found.
+    :raises KeyError: If the section is not found in the configparser object.
+    """
+
+    if not sc_settings.has_section(section_name):
+        raise KeyError(f"{section_name} section not found in {_SOURCE_CONTROL_INI_RELATIVE_PATH}")
```
baxeaz, line 52: Is there anything we should be asking them to do here to try to fix it?

src/deadline/unreal_perforce_utils/unreal_source_control.py
```diff
@@ -0,0 +1,179 @@
...
+    :param option_name: The name of the option to check for within the section.
+    :return: The configparser object if the option is found in the specified section.
+    :raises KeyError: If the option is not found in the specified section of the configparser object.
+    """
+
+    try:
+        sc_settings.get(section_name, option_name)
+    except configparser.NoOptionError:
+        raise KeyError(
+            f"{section_name} section does not contain '{option_name}' setting "
```
baxeaz, line 76: Same question as above here, should they be reenabling perforce, or editing it manually, or are they basically in some bad state here that we don't really know the right way to resolve?

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
```diff
@@ -752,6 +752,20 @@ def _update_environments_settings_from_mrq_job(
             if override_environment:
                 env.variables = override_environment.variables.variables
 
+    @staticmethod
+    def _get_project_path_relative_to_workspace_root(workspace_root: str) -> str:
+        workspace_root = workspace_root.replace("\\", "/")
+        unreal_project_path = common.get_project_file_path().replace("\\", "/")
+        if not unreal_project_path.startswith(workspace_root):
```
baxeaz: It would be an edge case for sure, but it looks like there might be a potential issue here if customers were to change the casing in their perforce settings to some part of the folder which had already been created.  Just for the startswith check it might be worth doing a lower() conversion as well?

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_custom_step_handler.py
karthikbekalp: Thanks!

karthikbekalp (commented):
Finished a review and didn't find any issues in particular other than the ones called out by @baxeaz . Also, the DCO has failed and as pointed out earlier will require a rebase.

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
alf-devnull: Ahhh, yeah, absolutely so. Will do!

src/deadline/unreal_perforce_utils/unreal_source_control.py
alf-devnull: Actually, its Just in case check. In general, the user will never change this file. It is enough for him to change the values ​​in the editor. And Unreal discover the connection properties automatically and check if the project under some workspace

src/deadline/unreal_perforce_utils/unreal_source_control.py
alf-devnull: As answered below, this just in case check. Sections may disappear only if user will change this file at runtime. The Unreal Editor overwrite it on the start/exit anyway with appropriate settings

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
alf-devnull: was thinking about it. Even if we check startswith with lower, we fail later on replace. For example if Unreal returns path C:/Perforce/Workspace1/Proj.uproject
Perforce root will be C:/Perforce/workspace1
its the same in startswith (with lower)
but replace will not work.

I think we should use some regex for that

src/deadline/unreal_perforce_utils/unreal_source_control.py
baxeaz: Perfect, thanks!

src/deadline/unreal_perforce_utils/unreal_source_control.py
baxeaz: Great, thanks!

karthikbekalp (approved):
It was a little hard to review with multiple commits and rebase. 

Anyway, thanks for your contribution. Approving this PR.

baxeaz (approved):

Issue #105, opened 2025-01-06
Bug: Multiple UE shots appearing as unique tasks all have same name, and render ordering appears random


andymanaws:
### Describe the bug

UE tasks are still all called the same thing - resulting in random(?) rendering of shots.

Can we :

- change the 'handler' column header to 'shot' - in other apps the 'handler' sections is called frames for example.
- ensure that the name of the shot is not 'render' but the actual shot name itself being rendered - e.g. shot0010. If Deadline Cloud uses an internal sort of the 'handler' to determine render order - e.g. I imagine thats fine for a numeric like typical frames - then we might need to prepend an chronological/execution number to the name to ensure the shots are rendered in chronological order, as the chronological order in UE is driven by the sequencer, and not the name of the shot itself - e.g. we probably need "shot01_shot0010", "shot02_shot0015", "shot03_shot0011",etc/

### Expected Behaviour

Shots are displayed in the 'handler' field with their name, so that we know which tasks are which shots, and the shots are executed in the order that they appear chronologically coming from the sequencer- so the first chronological shot is rendered first, then the second ,etc. 

### Current Behaviour

All the shots are called the same, and the render order appears random ( either the tasks are oredered wrongly in the DCM or the workers are picking up a random order).

### Reproduction Steps

Submit a job with multiple shots.

### Environment

At minimum:
1. Operating system: Windows Server 2025
2. Version of Unreal Engine: 5.5
3: Version of this package: 
deadline                         0.49.3
deadline-cloud-worker-agent      0.27.4
4. If this is from a version installed by the Deadline Cloud Submitter installer, then what version of the submitter installer?

Please share other details about your environment that you think might be relevant to reproducing the bug.

baxeaz:
Good suggestions, will take a look, thanks!

baxeaz:
Partly fixed by #132 but leaving this open for the naming improvements

PR #132, opened 2025-03-11
feat: Switching to sorting tasks by ChunkId in DCM

baxeaz:
Tasks in DCM appeared sorted in a random order because they're sorted based on the "handler" column by default which doesn't currently contain any useful information for sorting.

This is mentioned in #105 however this is not the complete solution discussed there - it won't include shot names in DCM.

Sort based on ChunkId which will better correlate to the order the work will be performed in.

Tasks will not appear to start in a random order - the first task in the list (ChunkId 0) should be the first to start.

Local testing:

chunkidsorted

Yes

No

No

----


crowecawcaw (approved):

karthikbekalp (approved):
Very interesting. I didn't know that the DCM worked in such a way.

PR #133, opened 2025-03-12
feat: Adding checks with warnings to the build_plugin script for comm…

baxeaz:
We've encountered a couple different configuration issues which required time spent on debugging - we'd like to try to prevent customers in the future from hitting these snags.

Start checking for common configuration issues in our build script and warn customers about the problems.

Less frustration for customers

Local testing with various configuration combinations

Yes

No

No

----

### DIFF
--- a/scripts/build_plugin.py
+++ b/scripts/build_plugin.py
@@ -11,12 +11,22 @@ import logging
 import shutil
 import os
 import subprocess
+import sys
 import tempfile
+from typing import Tuple, Optional
 
 DEFAULT_UE_INSTALL_ROOT = "C:\\Program Files\\Epic Games"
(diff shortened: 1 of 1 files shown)


crowecawcaw (approved):
Nice! Do we want to fail the build or is showing warnings sufficient?

joel-wong-aws (approved):

baxeaz:
> Nice! Do we want to fail the build or is showing warnings sufficient?

Good question!  I was a bit torn on this.  Technically long paths not being enabled doesn't 100% mean you're going to fail, and if you saw either warning (And your build didn't fail) you should be able to go resolve the problem before kicking off a job.  So I went with warnings but maybe down the line we'll want to revisit if folks miss the warnings.

PR #137, opened 2025-03-15
test: Created an e2e test which can run through hatch which wraps the…

baxeaz:
We want to be able to run end to end tests that validate that changes to code still build, on the submitter side still allow us to submit a job, and on the adapter side still successfully render the job.

Create additional test support and an e2e test in pytest which can be run through hatch run e2e which uses our build script to build and install the binaries as well as both the submitter and adapter, creates a temporary project from an MRQ compatible template and configures it with the necessary plugins, and runs the CreateJob test from the command line.

Future changes will create deadline resources (Farm/queue/fleet/etc) for the purposes of the test, start a local worker agent using those resources, submit to that queue, and validate outputs.

Much easier to validate changes.

Ran test successfully against local worker agent, observed job completed successfully.

e2ecreatejob

Yes

No

----

### DIFF
--- a/hatch.toml
+++ b/hatch.toml
@@ -5,7 +5,8 @@ pre-install-commands = [
 
 [envs.default.scripts]
 sync = "pip install -r requirements-testing.txt"
-test = "pytest --cov-config pyproject.toml {args:test}"
+test = "pytest test/ --cov-config pyproject.toml --ignore=test/end_to_end {args}"
+e2e = "pytest test/end_to_end --no-cov {args}"
 typing = "mypy {args:src test}"
 style = [
   "ruff check {args:.}",
--- a/scripts/__init__.py
+++ b/scripts/__init__.py
@@ -1,0 +1,1 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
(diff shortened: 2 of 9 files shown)


crowecawcaw (approved):
Nice! It'd be great to have updates to docs as well to explain why we have this new test suite and how to run them.

baxeaz:
> Nice! It'd be great to have updates to docs as well to explain why we have this new test suite and how to run them.

Definitely!  Will add with the next PR

karthikbekalp (approved):
Nice thanks for adding these tests. Although the Semantic PR check has failed, can you take a look at that? 

Is it because the commit has `feat` in it even though its a test?

baxeaz:
> Nice thanks for adding these tests. Although the Semantic PR check has failed, can you take a look at that?
> 
> Is it because the commit has `feat` in it even though its a test?

No that was just me trying to debug it initially - it turns out there's an open issue with semantic prs (https://github.com/Ezard/semantic-prs/issues/822) It looks like I might be able to work around it by removing most of the commit message but was going to give it a bit today to see if it gets resolved first

PR #140, opened 2025-04-24
test: Adding support code for new e2e tests

baxeaz:
We want to add more thorough end to end tests which can validate that jobs can still be run successfully as changes are being made to the code - both by customers iterating locally or as part of an automated review of our PRs.

Add support for creating, using, and cleaning up test resources.

Currently no impact - this is the first piece which is just support code, the additional changes including tests using these features are in upcoming PRs.

Ran the tests and cleanup script successfully.

Yes with the tests I have coming I verified they passed and was able to download the output.

e2e test changes will be documented in an upcoming PR.

No - this change is intended to add test support only.

----

### DIFF
--- a/scripts/clean_test_resources.py
+++ b/scripts/clean_test_resources.py
@@ -1,0 +1,401 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+"""
+Cleanup script for Deadline Cloud test resources.
+
+This script identifies and removes test resources created during Deadline Cloud for Unreal Engine tests,
+including queue-fleet associations, queues, and fleets.
+"""
+
+import argparse
+import boto3
+import subprocess
+import time
+import logging
+import sys
+
+# Configure logging
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s - %(levelname)s - %(message)s",
+    handlers=[logging.StreamHandler(sys.stdout)],
+)
+logger = logging.getLogger(__name__)
+
+# Constants
+DEADLINE_UNREAL_QUEUE_TEST_ROLE = "DeadlineUnrealQueueTestRole"
+DEADLINE_UNREAL_FLEET_TEST_ROLE = "DeadlineUnrealFleetTestRole"
+DEADLINE_UNREAL_TEST_QUEUE_NAME = "deadline-unreal-test-queue"
+DEADLINE_UNREAL_TEST_FLEET_NAME = "deadline-unreal-test-fleet"
+DEADLINE_UNREAL_TEST_FARM_NAME = "deadline-unreal-test-farm"
+
+
+def wait_for_qfa_stopped_status(deadline_client, farm_id, queue_id, fleet_id, max_wait_seconds=60):
+    """
+    Wait for a queue-fleet association to reach STOPPED status.
+
+    Args:
+        deadline_client: Boto3 Deadline client
+        farm_id: The farm ID
+        queue_id: The queue ID
+        fleet_id: The fleet ID
+        max_wait_seconds: Maximum time to wait in seconds
+
+    Returns:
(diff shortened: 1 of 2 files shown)


test/end_to_end/conftest.py
```diff
@@ -65,35 +285,484 @@ def add_plugins_to_project(project_path: str, plugins: list[str]):
...
+            logger.info(f"Using current execution role: {current_role_arn}")
+        else:
+            logger.warning(
+                "Not running as an IAM role, tests may fail if permissions are insufficient"
+            )
+
+        return current_role_arn
+
+
+def wait_for_job_state(
```
crowecawcaw, line 413: We have waiters for job create complete if that helps: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/deadline.html#waiters

We should probably expand our waiters to include any job state.

test/end_to_end/conftest.py
```diff
@@ -104,10 +773,979 @@ def create_readonly_test_project():
...
+    Fixture that provides an S3 client.
+
+    Args:
+        session: The boto3 session
+
+    Returns:
+        An S3 client
+    """
+    client = session.client("s3", region_name=TEST_TARGET_REGION)
+    logger.info(f"Created S3 client for region {TEST_TARGET_REGION}")
```
crowecawcaw: Nit: unnecessary logging

scripts/clean_test_resources.py
```diff
@@ -0,0 +1,399 @@
...
+
+# Constants
+DEADLINE_UNREAL_QUEUE_TEST_ROLE = "DeadlineUnrealQueueTestRole"
+DEADLINE_UNREAL_FLEET_TEST_ROLE = "DeadlineUnrealFleetTestRole"
+DEADLINE_UNREAL_TEST_QUEUE_NAME = "deadline-unreal-test-queue"
+DEADLINE_UNREAL_TEST_FLEET_NAME = "deadline-unreal-test-fleet"
+DEADLINE_UNREAL_TEST_FARM_NAME = "deadline-unreal-test-farm"
+
+
+def wait_for_qfa_stopped_status(deadline_client, farm_id, queue_id, fleet_id, max_wait_seconds=60):
```
crowecawcaw, line 32: Instead of manually creating resource and polling until they settle, what about making a CFN stack of required test resources, creating it at test setup, and deleting it for cleanup? Should be less code, more reliable, and more familiar to read and update.

test/end_to_end/conftest.py
baxeaz: Yeah that would be super helpful imo!  I could split the current code out and use the job create complete waiter for that purpose specifically and use this code for other states, I'm not sure if that's cleaner overall

test/end_to_end/conftest.py
baxeaz: Will remove!

scripts/clean_test_resources.py
baxeaz: So in my own testing I just reused my existing farm which was nice to be able to do - When iterating on the resources not having to do the extra step of creating/tearing down a farm.  Also farms I think are default limit of 5, so not eating one of those up for tests was also nice.  And honestly in general I really found it more convenient to not need to fight with/wait for CloudFormation and deal with all of the various rollbacks and such.

crowecawcaw: One killer CFN feature is the `disable-rollback` option which will pause a stack if it hits an error instead of rolling back. Then you can either update it again to move forward from where you left off. Makes iterating a lot faster! Doesn't need to be used here, but just a useful option.

baxeaz: Good to know about for sure!

crowecawcaw (approved):

karthikbekalp (approved):

PR #141, opened 2025-04-30
test: Adding an e2e test which will launch the worker agent, connect …

baxeaz:
…to our test queue, process a created test job and watch for the successful output.

We want to have more thorough end to end tests to validate changes being made to the plugin - whether in the submitter or adaptor - function correctly.

Create an end to end test which launches the worker agent locally and watches for a job created as part of the test to be run successfully.

Quicker and easier to validate changes in development and eventually as part of PR submission.

Ran test locally with hatch run e2e -s, verified tests pass, observed output.

Yes, this test runs a render job.

Yes

No


src/unreal_plugin/Content/Python/settings.py
```diff
@@ -73,49 +73,68 @@ def refresh_from_default_profile(self):
...
+            self.work_station_configuration.farm.job_attachment_filesystem_options = (
+                config.get_setting("defaults.job_attachments_file_system")
             )
 
-        self.work_station_configuration.farm.job_attachment_filesystem_options = config.get_setting(
-            "defaults.job_attachments_file_system"
-        )
+            self.work_station_configuration.general.auto_accept_confirmation_prompts = (
+                True if config.get_setting("settings.auto_accept") == "true" else False
+            )
```
crowecawcaw: Nit: simpler as just `config.get_setting("settings.auto_accept") == "true"`

src/unreal_plugin/Source/UnrealDeadlineCloudService/Private/Tests/Integration/DeadlinePluginTest_CreateJobTest.cpp
```diff
@@ -118,6 +141,167 @@ class RestoreQueueCommand : public IAutomationLatentCommand
...
+                        }
+                        else
+                        {
+                            UE_LOG(LogCreateJobTest, Display, TEXT("Queue value unchanged (already '%s'), skipping update"), *Value);
+                        }
+                    }
+                }
+                else
+                {
+                    UE_LOG(LogCreateJobTest, Warning, TEXT("Failed to split pair '%s' on '='"), *KeyValuePairs[i]);
```
crowecawcaw, line 277: Should we fail here?

src/unreal_plugin/Content/Python/settings.py
```diff
@@ -240,12 +259,64 @@ def on_settings_modified(self, property_name):
...
+                # Check if the value is actually changing
+                current_farm_id = config.get_setting("defaults.farm_id")
+                if current_farm_id != farm_value:
+                    logger.info(f"Farm value appears to be an ID, setting directly: {farm_value}")
+                    config.set_setting("defaults.farm_id", farm_value)
+                else:
+                    logger.info(f"Farm ID unchanged (already {farm_value}), skipping update")
+            else:
+                # It's a name, look it up
+                farm = self.find_farm_by_name(farm_value)
```
crowecawcaw: Is this part of the test or the submitter? We don't usually use farm names for lookups in other places. Is it just to make testing simpler? The profile should only contain farm IDs AFAIK.

src/unreal_plugin/Content/Python/settings.py
baxeaz: Fixed!

src/unreal_plugin/Source/UnrealDeadlineCloudService/Private/Tests/Integration/DeadlinePluginTest_CreateJobTest.cpp
baxeaz: Hmmm sort of hard to know what case we'd arrive here at.  If we did we'd likely apply the same logic to a bunch of other cases as well - I tend to prefer warning over just failing for things that aren't obviously going to cause things to fail later.  I think I'd rather wait and see if this becomes a problem area, we could always start failing faster if it does

src/unreal_plugin/Content/Python/settings.py
baxeaz: This was a good call out!  On the submitter in unreal we actually _do_ use names and then go lookup the ID based on the name.  Having this extra code here to support setting directly by ID is kind of going around the normal flow when really we want to test it as much as possible.  I've removed the id setting code and though the test communicates the queue/farm by id, we now call our underlying find_name_by_id methods in order to figure out what to set to.  In theory there could be naming collisions, but our test code already tries to reuse existing resources based on name, so this new approach seems reasonable to me.

crowecawcaw: How does the submitter itself use the farm name? Sounds like a inconsistent behavior with other integrations. Also sounds like it's not really part of this PR so not blocking, just curious.

crowecawcaw (approved):

karthikbekalp (approved):

karthikbekalp:
nit: It would be nice to add a step to run the test when raising a PR for the repo? https://github.com/aws-deadline/deadline-cloud-for-unreal-engine/blob/mainline/.github/PULL_REQUEST_TEMPLATE.md

PR #149, opened 2025-05-28
fix: Updating docs to reflect new version compatibility

baxeaz:
Due to a crash issue we've been experiencing in version 5.5 we've not been able to confirm we support 5.5  After investigating it appears to have been due to a few different issues.  In DirectX 11 there's an issue which has been fixed with a commit listed [here](https://issues.unrealengine.com/issue/UE-276282) which is expected to come out in 5.6 but offers a workaround for DirectX 11 users.  For 5.5 there was a fix pushed in March in 5.5 which appears to resolve the issue for DirectX 12.  On the Deadline Cloud side we can now say we support version 5.5 as the only remaining fix is one we've identified on Unreal's side which has a couple potential workarounds.

Additionally, recent changes in openjd-model which require a context object when initializing FormatString derived classes had broken the plugin when building against latest - switched to instantiating using the parse_model method which will pull the default context from the underlying classes.

Cleaned up some additional linting and build warnings.

Test, confirm, update docs

Customers can feel confident using 5.5.  The number of users still on DirectX 11 is reported by Epic as "low".

Docs only, but ran a render job to confirm we could say we support 5.5

Yes

It is docs

No

----

### DIFF
--- a/SETUP_SUBMITTER_CMF.md
+++ b/SETUP_SUBMITTER_CMF.md
@@ -10,7 +10,7 @@ These instructions are updated along with the corresponding code and scripts fai
 
 If you’re setting up on a brand new Windows EC2 Instance as your submitter, a g5.2xlarge instance with 200 GB of storage will likely be reasonable minimum:
 
-1. Download the Epic Installer and install a version of Unreal between versions 5.2 and 5.4. With 5.5 a bug has been observed that may prevent headless rendering on Deadline Cloud - we recommend using version 5.4.x at latest currently.
+1. Download the Epic Installer and install a version of Unreal between versions 5.2 and 5.5.  Note that on version 5.5 with DirectX 11 there's a crash bug which can affect projects rendered using the Deadline Cloud plugin which has been fixed in Unreal's source and can be tracked [here](https://issues.unrealengine.com/issue/UE-276282).  Projects in Deadline Cloud should use DirectX 12 with UE 5.5.
 1. NVIDIA GRID drivers - Follow Windows instructions - https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/install-nvidia-driver.html#nvidia-GRID-driver
 
 ## Windows Long Paths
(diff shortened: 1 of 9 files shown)


SETUP_SUBMITTER_CMF.md
```diff
@@ -10,7 +10,7 @@ These instructions are updated along with the corresponding code and scripts fai
 
 If you’re setting up on a brand new Windows EC2 Instance as your submitter, a g5.2xlarge instance with 200 GB of storage will likely be reasonable minimum:
 
-1. Download the Epic Installer and install a version of Unreal between versions 5.2 and 5.4. With 5.5 a bug has been observed that may prevent headless rendering on Deadline Cloud - we recommend using version 5.4.x at latest currently.
+1. Download the Epic Installer and install a version of Unreal between versions 5.2 and 5.5.  Note that on version 5.5 with DirectX 11 there's a crash bug which can affect projects which has been fixed in Unreal's source and can be tracked [here](https://github.com/EpicGames/UnrealEngine/commit/e1b60c8c77dd568e51f8c61011a54a54e0c04e84).  Projects in Deadline Cloud should use DirectX 12 with UE 5.5.
```
komapa: `[...] there's a crash bug that affects deadline cloud rendering plugin and which has been fixed [...]`

baxeaz: Hmm it's not really specific to the deadline cloud plugin though - the requirements as I understand it are that you're using -renderoffscreen (which we do) and rendering with DirectX 11 (Which will depend on your settings/capabilities, but by default Unreal 5 should prefer 12).  Either way I'm glad you pointed this out though because I think I'm not linking to the correct tracking issue which is https://issues.unrealengine.com/issue/UE-276282 and originally came up from users of the PixelStreaming plugin

baxeaz: Updated the phrasing, let me know if that works!

karthikbekalp (approved):

crowecawcaw (approved):

SETUP_SUBMITTER_CMF.md
komapa: Reads well, thank you!

PR #167, opened 2025-07-16
chore: add native dependency psutil to deps bundle

sakshie95:
Fixes: N/A

We added a new dependency psutil to deadine-cloud https://github.com/aws-deadline/deadline-cloud/blob/e5176598cc859584217eaea00d5e6a14a40070d5/pyproject.toml#L47 for a new feature in experimental. This dependency has native components and needs to be added to deps bundle for every installer

Added psutil to native dependencies for the feature to be supported.

Bundling required dependencies for every submitter installer. For unreal there is no individual installer but we're doing this for consistency with other DCCs

``hatch build``, ``pip install`` and ``hatch run test``. Couldn't build the installer to verify for this case

Yes
```
-------------------------------------------------------------------------------------------------------------------------------------
TOTAL                                                                                      2236    731    538     41    63%
Coverage HTML written to dir build/coverage
Coverage XML written to file build/coverage/coverage.xml
Required test coverage of 43.0% reached. Total coverage: 63.45%
================================================================================================================= slowest 5 durations ==================================================================================================================
10.07s call     test/deadline_adaptor_for_unreal/unit/UnrealAdaptor/test_adaptor.py::TestUnrealAdaptor_on_cleanup::test_raises_if_unreal_not_running
9.24s call     test/deadline_adaptor_for_unreal/unit/UnrealAdaptor/test_adaptor.py::TestUnrealAdaptor_on_start::test_init_data_wrong_schema
0.15s call     test/test_copyright_headers.py::test_copyright_headers
0.11s call     test/deadline_adaptor_for_unreal/unit/UnrealAdaptor/test_adaptor.py::TestUnrealAdaptor_on_start::test_unreal_init_timeout
0.11s call     test/deadline_submitter_for_unreal/unit/test_submitter.py::TestUnrealSubmitter::test_submit_jobs
=========================================================================================================== 269 passed, 1 skipped in 41.95s ============================================================================================================
```

No, not required for this use-case. Tested an installer e2e for Maya where an individual installer was buildable.

Not required, I've followed the same process as xxhash

No

----


Cherie-Chen:
If this dependency isn't essential, we should avoid adding it solely for consistency reason. Unnecessary dependencies increase the project size and introduce additional risks, such as potential security vulnerabilities (CVEs).

sakshie95:
Discarding change because we learned that the ``deps_bundle`` is actually deprecated for this package. This is dead code which will be tested & released separately

PR #175, opened 2025-07-25
fix: Parameters definitions are empty in MRQ UI when you first time create the job. But if you reset the data asset, then they appear

slavauz:

Parameters definitions are empty in MRQ UI when you first time create the job. But if you reset the data asset, then they appear

Override `PostInitProperties` in `UMoviePipelineDeadlineCloudExecutorJob`. In that method, trigger the application of default values from the data asset. This allows the default job values to be displayed in the MRQ UI immediately after the job is created

Display default parameters definitions in MRQ UI when you first time create the job

QA Test: Create a new job, switch presets, and make sure the default parameter definitions are updated accordingly

Submitted a test job to the CMF instance. The job succeeded

Bug fix which doesn't require documentaion

No

----


Cherie-Chen (approved):

baxeaz (approved):

komapa:
Thank for the fix. Would it be possible to get a bugfix release for 0.6.0 so we do not have to install from mainline to get this pretty important UI elements?

baxeaz:
Hi @komapa! I think we're in a good spot for a patch release - we'll take a look today.

Issue #154, opened 2025-06-16
Bug: Contains empty elements or more than 50 of them with UE 5.5.4


komapa:
### Describe the bug

One error that we get when submitting with UE 5.5.4 is the following: `Contains empty elements or more than 50 of them`

We checked the logs, checked elements and tried to solve it ourselves, but don't know where to look anymore. There is nowhere we can find what that means, or which elements are empty. We did everything like before (on EU 5.4.4), so not sure if this is new to UE 5.5 or not.

![Image](https://github.com/user-attachments/assets/51591cd8-d8cc-48fa-934a-4e81f6cb75ab)

### Expected Behaviour

Be able to submit the job

### Current Behaviour

Error: `Contains empty elements or more than 50 of them`

### Reproduction Steps

N/A

### Environment

Windows 11, with Unreal Engine 5.5.4 and deadline-cloud-for-unreal-engine installed from the `main` branch.

komapa:
On the Content side, the update redirector references have been completed, and everything without references should now be clean. All materials and textures that were moved from A to B have been checked; everything I used in the Outliner are existing, valid files with proper references from the Content folder.
There were some file mentions in the log that were still being referenced during array counting, but they no longer had any references in the Unreal project itself. I located those in the Content folder and deleted them "physically" from the Content Browser.
The Datasmith files are still present in the project, but their link to the assets (materials/meshes) has been broken from the asset user data.
Everything was done the same way as before, theoretically. "So, when I place the sequence into the Movie Render Queue and select auto-detect for the input files, it calculates (I have around 5k assets). That's when the following error appears.
For context: we import assets into Unreal from 3ds Max using Datasmith (this has been our method even before, while using 5.4 as well).

baxeaz:
Thank you for the submission @komapa !  I hadn't seen this one before but am able to reproduce the UI error in Meerkat when I turn on Show Autodetected.  It doesn't seem to block me from submitting though - it does seem to be blocking you though?  If you don't turn on show auto detected are you able to submit or do you end up getting the same error?

komapa:
> Thank you for the submission [@komapa](https://github.com/komapa) ! I hadn't seen this one before but am able to reproduce the UI error in Meerkat when I turn on Show Autodetected. It doesn't seem to block me from submitting though - it does seem to be blocking you though? If you don't turn on show auto detected are you able to submit or do you end up getting the same error?

[...]

PR #176, opened 2025-07-25
fix: Error message in  field: Contains empty elements or more than 50 of them with UE 5.5.4 (#154)

slavauz:

`Auto Detected Files` field includes a default validation that limits the list to a maximum of 50 elements. In Unreal, this limit is consistently exceeded. The validation also checks for empty elements, which can result in a misleading error message causing the user to think there are empty entries in the list, even when there aren't

- Remove 50 elements limit check for `Auto Detected Files`
- Add separate validations for the maximum number of elements and for empty elements

Updated paths list validation logic in UI

QA UI behavior tests for paths lists

Submitted test job succeeded

Doesn't require documentation update

No

----


Cherie-Chen (approved):
Tested this change in UE 5.6, looks good.

Before this code change, misleading error message in UI,
Screenshot 2025-07-25 at 12 55 55 PM

After this change, 
Screenshot 2025-07-25 at 12 41 50 PM

baxeaz (approved):

PR #180, opened 2025-07-31
fix: Autodetect Project Plugins

alf-devnull:

Need to implement Project plugins autodetection system similar to the Project Assets autodetection system.

Parse .uproject and .uplugins files to build the Plugins list that used in this Project

List of the Project plugins automatically attached to asset references of the OpenJob.

Submit the simple Render OpenJob on the project with some plugins enabled. Autodetected directories must contain these plugins

Yes

Yes, docstrings

No
----

### DIFF
--- a/src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
+++ b/src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
@@ -416,42 +416,44 @@ class UnrealOpenJob(UnrealOpenJobEntity):
         return asset_references
 
     @staticmethod
+    def get_plugins(path: str):
+        unreal_plugins: list[dict] = []
+        pattern = os.path.join(path, "**", "*.uplugin")
+
+        for uplugin in glob.iglob(pattern, recursive=True):
+            real_path = Path(uplugin).resolve(strict=True)
+            try:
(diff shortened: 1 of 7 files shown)


src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
```diff
@@ -403,6 +403,55 @@ def get_asset_references(self) -> AssetReferences:
...
+        plugins = get_plugins(plugins_dir_full)
+
+        for plugin in plugins:
+            if plugin["name"] == "UnrealDeadlineCloudService":
+                continue
+
+            is_enable = project_plugins_info.get(plugin["name"], plugin["enabled_by_default"])
+
+            if is_enable:
+                result.input_directories.add(plugins_dir_full + plugin["folder"])
```
Cherie-Chen: NIT: use `os.path.join()` instead of string concatenation can help preventing invalid file paths.

Cherie-Chen (approved):

Cherie-Chen (changes requested):
Could you please include test coverage for this feature? Unit tests would be great to start with, and if feasible, E2E tests would be valuable additions as well.

Cherie-Chen (approved):

karthikbekalp (approved):

PR #188, opened 2025-08-21
docs: Add troubleshooting guide in DEVELOPMENT.md

Cherie-Chen:
Fixes: NA

Add troubleshooting guide in DEVELOPMENT.md

Add troubleshooting guide in DEVELOPMENT.md

Document the error messages and solutions to help others resolve similar issues

Yes

No. No functional change.

Yes

No

----

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -89,7 +89,7 @@ The Deadline Cloud plugin's Unreal Automation Tests can be run from within Unrea
 To test out any significant changes it's useful to submit a test render following [this guide](https://github.com/aws-deadline/deadline-cloud-for-unreal-engine/blob/mainline/SETUP_SUBMITTER_CMF.md#submit-a-test-render-optional)
 
 
-### Building the docs
+## Building the docs
 
 1. Install python requirements for building Sphinx documentation
    ```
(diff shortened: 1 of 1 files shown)


DEVELOPMENT.md
```diff
@@ -114,10 +114,53 @@ To test out any significant changes it's useful to submit a test render followin
...
+      ],
+      "Resource": "arn:aws:logs:*:*:log-group:/aws/deadline/*"
+    }
+  ]
+}
+```
+
+### Unreal Engine Version Mismatch
+
+Error: The package '/Temp/UnrealDeadlineCloudService/RenderJobManifests/###' was saved with an older version which is not backwards compatible with the current process
```
crowecawcaw, line 160: Is there a way we can detect these errors earlier and surface them in a nicer error message? I saw someone else get stuck on this version mismatch issue recently as well.

crowecawcaw (dismissed):

DEVELOPMENT.md
Cherie-Chen: Yes, we're adding automatic Unreal Engine version detection. CMF customers may still run into this problem if their worker nodes don't have the correct version installed. It's still worth documenting.

DEVELOPMENT.md
```diff
@@ -114,10 +114,53 @@ To test out any significant changes it's useful to submit a test render followin
...
+
+Root Cause: Version mismatch between the Unreal Engine version used to submit the job and the version running on the worker node.
+
+Solutions:
+   - Resubmit the job using the Unreal Engine version that matches the worker node
```
baxeaz: We should probably call out the SMF/conda package version case here as well - even once we have the auto detection it will be possible for customers to override it.  Something like - "On Service Managed Fleets - Ensure the Conda package version selected matches your project's version of Unreal Engine

Cherie-Chen: Got it. Will update the PR

baxeaz (approved):

crowecawcaw (approved):

PR #196, opened 2025-08-30
feat: Set right UE version to CondaPackages parameter

alf-devnull:

It's necessary that CondaPackages parameter should contain same UE version as the version of the project from job is submitted. Therefore the default value define in Template should be updated on submit if its differs

Add validation of CondaPackages parameter. 
- If CondaPackages parameter not in OpenJob parameters - skip validation.
- Otherwise check if its value contains unrealengine=x.x (Regex)
   - If contains compare with current UE version of the Project
      - If equals, continue submit. Raise an Error otherwise
   - If its not there, build with current version

Job can't be submitted with wrong unreal engine conda package version

Submit Job without CondaPackages parameter - validation skipped
With empty CondaPackages parameter - unrealengine=x.x were added (5.4)
With CondaPackages=unrealengine=5.3 - raise an error (Project was on 5.4)
With CondaPackages=unrealengine=5.4 - validation passed (Project was on 5.4)

Yes

Yes

Yes

No
----

### DIFF
--- a/src/deadline/unreal_submitter/exceptions.py
+++ b/src/deadline/unreal_submitter/exceptions.py
@@ -31,6 +31,12 @@ class OpenJobIsMissingError(DeadlineCloudSubmitterException):
     pass
 
 
+class InvalidUEVersionInCondaPackageParameter(DeadlineCloudSubmitterException):
+    """Raised when Conda Package parameter is contains invalid ue version"""
+
+    pass
+
+
 class RenderArgumentsTypeNotSetError(DeadlineCloudSubmitterException):
     """Raised when the render arguments type is not set"""
 
--- a/src/deadline/unreal_submitter/exceptions.py
+++ b/src/deadline/unreal_submitter/exceptions.py
@@ -7,8 +7,8 @@ class DeadlineCloudSubmitterException(Exception):
     pass
 
 
-class NoUINotification:
-    """Marker mixin: Exceptions that shouldn't trigger UI notifications"""
+class UserException:
+    """Marker mixin: Exceptions that do not trigger UI notifications and are raised by user choice."""
 
     pass
 
@@ -37,15 +37,13 @@ class OpenJobIsMissingError(DeadlineCloudSubmitterException):
     pass
 
 
-class InvalidUEVersionInCondaPackageParameter(DeadlineCloudSubmitterException):
-    """Raised when Conda Package parameter contains an invalid UE version"""
+class UEVersionParseError(DeadlineCloudSubmitterException):
+    """Raised when the current Unreal Engine version string cannot be parsed (expected 'x.y')."""
 
     pass
 
 
-class InvalidUEVersionInCondaPackageParameterNoUI(
-    InvalidUEVersionInCondaPackageParameter, NoUINotification
-):
+class UserCancelledSubmissionMissMatchedUEVersion(DeadlineCloudSubmitterException, UserException):
     """Raised when Conda Package parameter contains an invalid UE version"""
 
     pass
--- a/src/deadline/unreal_submitter/exceptions.py
+++ b/src/deadline/unreal_submitter/exceptions.py
@@ -32,7 +32,7 @@ class OpenJobIsMissingError(DeadlineCloudSubmitterException):
 
 
 class InvalidUEVersionInCondaPackageParameter(DeadlineCloudSubmitterException):
-    """Raised when Conda Package parameter is contains invalid ue version"""
+    """Raised when Conda Package parameter contains an invalid UE version"""
 
     pass
 
--- a/src/deadline/unreal_submitter/exceptions.py
+++ b/src/deadline/unreal_submitter/exceptions.py
@@ -7,6 +7,12 @@ class DeadlineCloudSubmitterException(Exception):
     pass
 
 
+class NoUINotification:
+    """Marker mixin: Exceptions that shouldn't trigger UI notifications"""
+
+    pass
+
+
 class ParametersAreNotConsistentError(DeadlineCloudSubmitterException):
     """Raised when OpenJD parameters/variables are not consistent"""
 
@@ -32,6 +38,14 @@ class OpenJobIsMissingError(DeadlineCloudSubmitterException):
 
 
 class InvalidUEVersionInCondaPackageParameter(DeadlineCloudSubmitterException):
+    """Raised when Conda Package parameter contains an invalid UE version"""
+
+    pass
+
+
+class InvalidUEVersionInCondaPackageParameterNoUI(
+    InvalidUEVersionInCondaPackageParameter, NoUINotification
+):
     """Raised when Conda Package parameter contains an invalid UE version"""
 
     pass
--- a/src/deadline/unreal_submitter/exceptions.py
+++ b/src/deadline/unreal_submitter/exceptions.py
@@ -43,7 +43,7 @@ class UEVersionParseError(DeadlineCloudSubmitterException):
     pass
 
 
-class UserCancelledSubmissionMissMatchedUEVersion(DeadlineCloudSubmitterException, UserException):
+class UserCancelledSubmissionMismatchedUEVersion(DeadlineCloudSubmitterException, UserException):
     """Raised when Conda Package parameter contains an invalid UE version"""
 
     pass
--- a/src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
+++ b/src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
@@ -558,9 +558,16 @@ class UnrealOpenJob(UnrealOpenJobEntity):
         # Compare versions
         if not template_ue_version == current_version:
             # Versions don't match
-            raise exceptions.InvalidUEVersionInCondaPackageParameter(
-                "CondaPackages Unreal Engine version mismatch"
+            result = unreal.EditorDialog.show_message(
+                "Version Mismatch Warning",
+                f"You are attempting to render a UE {current_version} project using UE {template_ue_version}.Do you wish to continue?",
+                unreal.AppMsgType.YES_NO
             )
+
+            if result != unreal.AppReturnType.YES:
+                raise exceptions.InvalidUEVersionInCondaPackageParameter(
+                    "CondaPackages Unreal Engine version mismatch"
+                )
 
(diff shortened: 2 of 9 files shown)


src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
```diff
@@ -496,6 +502,68 @@ def create_job_bundle(self):
...
+
+        template_ue_version = ue_version_match.group(1)
+        logger.info(f"Template specifies Unreal Engine version: {template_ue_version}")
+
+        # Compare versions
+        if not template_ue_version == current_version:
+            # Versions don't match
+            raise exceptions.InvalidUEVersionInCondaPackageParameter(
+                "CondaPackages Unreal Engine version mismatch"
+            )
```
Cherie-Chen: Please change the error message. Something like : 
```
Version Mismatch Warning: You are attempting to render a UE {current_version} project using UE {template_ue_version}.

Do you wish to continue?'.
```

Cherie-Chen (changes requested):

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
```diff
@@ -648,6 +716,42 @@ def mrq_job(self, value):
         if self._name is None:
             self._name = self._mrq_job.job_name
 
+    def get_mrq_template_object(self) -> dict:
```
Cherie-Chen: This is not being used anywhere. 
This maybe part of PR189 and was removed already. Please double check.

Cherie-Chen:
Thank you for looking into this. This feature will be extremely helpful for our users.

Proposed workflow for this feature:
- Auto-Detection: Automatically detect user's current UE version
- Version Handling:
   - Override 'CondaPackages' default UE version with user's current version
   - Display a warning popup if submission target version differs from current version
   - Allow users to submit jobs with mismatched UE versions if they choose to proceed

This will provide maximum flexibility while maintaining proper warnings for users.

Cherie-Chen:
Formatting issue detected in unreal_open_job.py by code scanning tool.

To resolve:
- View issues: Run `hatch run lint`
- Auto-fix formatting: Run `hatch run fmt`

Cherie-Chen:
Hey, I've noticed something odd with the UE version detection. I'm running UE 5.5 (though I have 5.4 and 5.6 installed too), but the `CondaPackages` parameter is still showing 5.6. Oddly enough, I'm not seeing any validation warnings either, which is strange because I definitely got validation errors in the previous revision.

Could you please take a look again?

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
```diff
@@ -500,6 +502,75 @@ def create_job_bundle(self):
...
+
+        template_ue_version = ue_version_match.group(1)
+        logger.info(f"Template specifies Unreal Engine version: {template_ue_version}")
+
+        # Compare versions
+        if not template_ue_version == current_version:
+            # Versions don't match
+            result = unreal.EditorDialog.show_message(
+                "Version Mismatch Warning",
+                f"You are attempting to render a UE {current_version} project using UE {template_ue_version}.Do you wish to continue?",
```
Cherie-Chen: NIT: add a space before "Do you wish to continue?"

src/deadline/unreal_submitter/exceptions.py
```diff
@@ -31,6 +31,12 @@ class OpenJobIsMissingError(DeadlineCloudSubmitterException):
     pass
 
 
+class InvalidUEVersionInCondaPackageParameter(DeadlineCloudSubmitterException):
+    """Raised when Conda Package parameter is contains invalid ue version"""
```
Cherie-Chen: NIT: "is contains invalid ue version" -> "contains an invalid UE version"

alf-devnull:
> Hey, I've noticed something odd with the UE version detection. I'm running UE 5.5 (though I have 5.4 and 5.6 installed too), but the `CondaPackages` parameter is still showing 5.6. Oddly enough, I'm not seeing any validation warnings either, which is strange because I definitely got validation errors in the previous revision.
> 
> Could you please take a look again?

Regarding the appearance of 5.6: This value always taken from YAML template and we don't overwrite its content since its our source of truth. But that really strange that there were no warnings.

I've got a question: which UE version resulted job has: 5.6 or 5.5?

Cherie-Chen:
> This value always taken from YAML template and we don't overwrite its content since its our source of truth. 

Setting a default value for the `CondaPackages` parameter in `unrealengine` was implemented as a temporary solution to enable SMF support. We should enhance user experience  by automatically detecting and populating the Unreal Engine version that the user is currently working with.

We still need to keep the default value for script-based job submissions, as it will be used when jobs are submitted programmatically.


BTW, pop-up warning dialog is working as expected in UE5.5 and UE5.6.

Cherie-Chen:
> > Hey, I've noticed something odd with the UE version detection. I'm running UE 5.5 (though I have 5.4 and 5.6 installed too), but the `CondaPackages` parameter is still showing 5.6. Oddly enough, I'm not seeing any validation warnings either, which is strange because I definitely got validation errors in the previous revision.
> > Could you please take a look again?
> 
> Regarding the appearance of 5.6: This value always taken from YAML template and we don't overwrite its content since its our source of truth. But that really strange that there were no warnings.
> 
> I've got a question: which UE version resulted job has: 5.6 or 5.5?

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
```diff
@@ -500,6 +502,65 @@ def create_job_bundle(self):
...
+            )
+            return
+
+        template_ue_version = ue_version_match.group(1)
+        logger.info(f"Template specifies Unreal Engine version: {template_ue_version}")
+
+        # Compare versions
+        if not template_ue_version == current_version:
+            # Versions don't match
+            result = unreal.EditorDialog.show_message(
```
karthikbekalp, line 585: Hmm, what would happen if the UE version of the project and conda package don't match. Would the jobs still succeed? 

If jobs don't succeed, then I wonder if it would be simpler to just dynamically add the package version and not provide customers with extra pop-ups? 

We would have to also update the compatibility in our [README ](https://github.com/aws-deadline/deadline-cloud-for-unreal-engine?tab=readme-ov-file#compatibility) to document this though because we support UE 5.2.1 but we only provide conda package above or equal to UE 5.4.

alf-devnull: If the Job submitted to CMF, CondaPackages parameter will  be ignored I believe. Therefore we decide to not force the fail of submit in that case and just notify about it.

I think we need to come up with single solution from the both side to be on the same page regarding this feature

Cherie-Chen:
Small change request. 

When submitting a job with mis-matched UE version, we first popup a warning asking user if they want to continue. When user click "No", we pop up another error. Please remove the 2nd pop up of error. 

https://github.com/user-attachments/assets/8153354f-25cd-4e56-85ae-24fe10009f58

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
```diff
@@ -500,6 +502,77 @@ def create_job_bundle(self):
...
+                "Version Mismatch Warning",
+                f"You are attempting to render a UE {current_version} project using UE {template_ue_version}. Do you wish to continue?",
+                unreal.AppMsgType.YES_NO,
+                unreal.AppReturnType.YES,
+            )
+
+            if result != unreal.AppReturnType.YES:
+                raise exceptions.InvalidUEVersionInCondaPackageParameterNoUI(
+                    "CondaPackages Unreal Engine version mismatch"
+                )
```
Cherie-Chen: I suggest modifying the check_conda_package_version function to return a boolean value. This would make the logic more straightforward:

- Return true to continue job submission
- Return false to stop the process

This approach would make the code more readable and easier to maintain.

alf-devnull: Adressed

src/deadline/unreal_submitter/exceptions.py
```diff
@@ -31,6 +37,18 @@ class OpenJobIsMissingError(DeadlineCloudSubmitterException):
     pass
 
 
+class UEVersionParseError(DeadlineCloudSubmitterException):
+    """Raised when the current Unreal Engine version string cannot be parsed (expected 'x.y')."""
+
+    pass
+
+
+class UserCancelledSubmissionMissMatchedUEVersion(DeadlineCloudSubmitterException, UserException):
```
Cherie-Chen: Spelling issue. `MissMatched` -> `Mismatched`

Cherie-Chen (dismissed):

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
```diff
@@ -500,6 +505,76 @@ def create_job_bundle(self):
...
+        )
+
+        if not conda_packages_param:
+            return True
+
+        current_version = UnrealOpenJob.get_current_ue_version()
+
+        conda_packages_value = conda_packages_param.get("value", "")
+        if not conda_packages_value:
+            conda_packages_param["value"] = f"unrealengine={current_version}"
```
baxeaz: I like that we're making a best effort to fix things, but will this still fail them without an adapter (unrealengine-openjd=0.x.y) version?  We probably don't need to get too fancy and can default to latest (\*.\*.\*)

src/unreal_plugin/Content/Python/job_library.py
```diff
@@ -49,6 +52,54 @@ def get_job_dependencies(self, mrq_job):
...
+                break
+
+        if not conda_packages_param:
+            return parameters
+
+        current_version = UnrealOpenJob.get_current_ue_version()
+
+        conda_packages_value = conda_packages_param.value
+        if not conda_packages_value:
+            conda_packages_param.value = f"unrealengine={current_version}"
```
baxeaz: Similar comment as above - I think this will still fail without the adapter package?

Cherie-Chen (approved):

baxeaz (approved):

PR #229, opened 2025-10-27
feat: Adding support for FramesPerTask to set how many frames of rend…

### LINKED ISSUE #191: Feature request: Ability to split single shot into multiple jobs by splitting frame range
### Describe the problem

Hey! I've been using deadline cloud for a very large Unreal path tracer based job. The project is comprised of one long 40,000 frame shot. Rendering it on one instance would take a couple weeks, and the instances run out of memory (RAM not VRAM for some reason), so I've been splitting it into many jobs by overriding the frame range in each job and making each job render around 500-1000 frames. This means setting up anywhere between 40 and 80 separate jobs and manually setting the frame range for each one.

I've been thinking of making a tool to help me automate this, but I'm wondering if this is something you'd consider implementing on your side?

### Proposed Solution

Ideally I would love to have a new field in the Job Shared Settings that would determine the number of instances I want to run the job on. This would then take the frame range of the shot, and split it equally among that number of jobs.

### Example Use Cases

As path tracer based unreal projects grow I think many people will need to split single shots across multiple machines without manually splitting jobs.

baxeaz:
Fixes: #191 

The integration only supported dividing up rendering sequences by shots - it didn't support an out of the box way of dividing up rendering jobs by frames.

Add a FramesPerTask parameter which, if set, controls the number of frames each task should attempt to render.  By default this will divide up the total frames in the level sequence selected for the job by FramesPerTask to figure out how many tasks should be created.  This can also be overridden with the Unreal "Use Custom Playback Range" control - selecting your own Custom Start and End frame.

Customers can render by ranges of frames instead of just shots.

New unit tests added which pass.
Manual testing.
E2E tests pass.

Yes

Yes

No
----

### DIFF
--- a/docs/source/docs_usage/use_cases/create_render_job.rst
+++ b/docs/source/docs_usage/use_cases/create_render_job.rst
@@ -58,6 +58,9 @@ Render Job data asset
    .. image:: ../../images/create_render_job_6.png
 
    a. Executable - Unreal executable name to launch on render node
+   #. FramesPerTask - Divide up the total number of frames to be rendered in the sequence by this value to
+      determine the total number of Deadline Cloud tasks to create, with each task aiming to render this number
+      of frames.
    #. ExtraCmdArgs - Additional CMD arguments to launch Unreal executable with
    #. ExtraCmdArgsFile - Specific file parameter where **ExtraCmdArgs** will be stored.
       Need to avoid **1024 chars limit** on **STRING** parameter. **Filled automatically during the submission**
--- a/docs/user_guide/create-perforce-render-job.md
+++ b/docs/user_guide/create-perforce-render-job.md
@@ -140,6 +140,7 @@ Set up an OpenJD Render Job that orchestrates the entire rendering workflow.
 | `PerforceWorkspaceSpecificationTemplate` | Perforce client spec with `{workspace_name}` token | ✅ | **Leave empty** - Auto-populated |
 | `MrqJobDependenciesDescriptor` | JSON file with MRQ dependencies for sync | ✅ | **Leave empty** - Auto-populated |
 | `ExtraCmdArgsFile` | File for extra args (avoids 1024 char limit) | ❌ | **Optional** - Use default for standard setups |
+| `FramesPerTask` | Number of frames to render per task | ❌ | **Optional** - Use default (0) to divide tasks by shots |
 | `ExtraCmdArgs` | Additional Unreal launch arguments | ❌ | **Optional** - Use default for standard setups |
 | `Executable` | Unreal executable name for render node | ❌ | **Configure** - Use default for standard setups |
 | `CondaPackages` | Conda packages needed to render the job | ❌ | **Configure** - Use default for standard setups |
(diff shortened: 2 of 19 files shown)


docs/source/docs_usage/use_cases/create_render_job.rst
```diff
@@ -58,6 +58,9 @@ Render Job data asset
    .. image:: ../../images/create_render_job_6.png
 
    a. Executable - Unreal executable name to launch on render node
+   #. FramesPerTask - Divide up the total number of fraems to be rendered in the sequence by this value to
```
crowecawcaw:
```suggestion
   #. FramesPerTask - Divide up the total number of frames to be rendered in the sequence by this value to
```

docs/user_guide/create-perforce-render-job.md
```diff
@@ -140,6 +140,7 @@ Set up an OpenJD Render Job that orchestrates the entire rendering workflow.
 | `PerforceWorkspaceSpecificationTemplate` | Perforce client spec with `{workspace_name}` token | ✅ | **Leave empty** - Auto-populated |
 | `MrqJobDependenciesDescriptor` | JSON file with MRQ dependencies for sync | ✅ | **Leave empty** - Auto-populated |
 | `ExtraCmdArgsFile` | File for extra args (avoids 1024 char limit) | ❌ | **Optional** - Use default for standard setups |
+| `FramsPerTask` | Number of frames to render per task | ❌ | **Optional** - Use default (0) to divide tasks by shots |
```
crowecawcaw:
```suggestion
| `FramesPerTask` | Number of frames to render per task | ❌ | **Optional** - Use default (0) to divide tasks by shots |
```

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job_step.py
```diff
@@ -473,10 +474,36 @@ def _get_chunk_ids_count(self) -> int:
...
+                )
+                logger.info(
+                    f"Output settings at submission has range {output_settings.custom_start_frame} - {output_settings.custom_end_frame} ({total_frame_range} frames)"
+                )
+            else:
 
+                total_frame_range = (
+                    level_sequence.get_playback_range().get_end_frame()
+                    - level_sequence.get_playback_range().get_start_frame()
+                )
```
crowecawcaw, line 497: Do we need to add 1? I'd expect frame range 1 -> 1 would render a single frame.

baxeaz: Unreal uses inclusive starts and exclusive ends for Sequencer/MRQ - https://dev.epicgames.com/documentation/en-us/unreal-engine/cinematic-workflow-tips-for-sequencer-in-unreal-engine#inclusive-and-exclusive-frames.  Makes the math a bit easier at least :)

docs/source/docs_usage/use_cases/create_render_job.rst
baxeaz: Fixed, thanks!

docs/user_guide/create-perforce-render-job.md
baxeaz: Fixed, thanks!

joel-wong-aws (approved):

Cherie-Chen (approved):

PR #269, opened 2026-03-03
docs: Add AGENTS.md for each meaningful subfolder

Cherie-Chen:
docs: Add AGENTS.md for each meaningful subfolder

Adopting (agents.md)[https://agents.md/] to give AI agents more context of each meaningful subfolder. 

Add `AGENTS.md` in each meaningful subfolder.

AI agents will have more context when working with this code base.

NA

No, there is no logic change

No, there is no logic change

This is a doc change

No

----


src/deadline/AGENTS.md
```diff
@@ -0,0 +1,41 @@
...
+- `unreal_perforce_utils` — Perforce workspace management and source control operations
+- `unreal_logger` — Logging utilities that bridge Python logging to Unreal's log system
+- `unreal_cmd_utils` — Unreal Engine command-line argument parsing and merging
+
+## Build & test
+
+- Build system: Hatch (`pyproject.toml` at repo root)
+- Python: >=3.9
+- Install dev: `pip install -e ".[dev]"` from repo root
+- Run tests: `pytest` from repo root (configured in `pyproject.toml`)
```
baxeaz: Is this just describing the underlying framework?  It looks like it's saying that's how to run tests, but that should be through "hatch run test" I think?

src/deadline/AGENTS.md
```diff
@@ -0,0 +1,41 @@
...
+- Coverage threshold: 65%
+- Linting: `ruff` (line-length 100), `mypy` (check_untyped_defs, Python 3.9)
```
baxeaz: Similar here, should this mention "hatch run lint" and below "hatch run fmt"?

test/AGENTS.md
```diff
@@ -0,0 +1,31 @@
+# AGENTS.md — test
+
+## Overview
+
+Test suite for the `deadline-cloud-for-unreal-engine` Python packages. Organized by source package with unit and end-to-end test directories.
+
+## Running tests
+
+- Run all tests: `pytest` from repo root
```
baxeaz: Here also, I think this should all be through hatch?

src/deadline/AGENTS.md
Cherie-Chen: You are totally right, 2 issues here,
1. These statements are incorrect
2. Content related to dev, build, test, deployment, and coding style does not belong in AGENTS.md files — that belongs in dev_power or development.md. Including it here creates duplication, clutters the file, and wastes tokens.
I've removed these from all AGENTS.md files.

test/end_to_end/AGENTS.md
```diff
@@ -0,0 +1,18 @@
...
+## Key files
+
+- `conftest.py` — Extensive fixture setup: AWS resource creation (farms, queues, fleets), Unreal Engine discovery, logging configuration. Uses `boto3` directly.
+- `test_create_job.py` — Tests for job creation
+- `test_worker_agent.py` — Tests for worker agent behavior
+
+## Important context
+
+- These tests require valid AWS credentials and a real Deadline Cloud environment
+- They are slow and resource-intensive — not part of the default `pytest` run
```
baxeaz: This one is maybe less of an issue - but it does kind of imply there's a "default pytest run" - maybe update it to say "not run as part of the default unit tests" if we don't want to call out the "hatch run test" command?

baxeaz (approved):

leon-li-inspire (approved):

PR #273, opened 2026-03-20
feat: Adding persistent perforce workspace support

baxeaz:
- Added WorkspaceInfoFile dataclass with load/save/lookup/register methods
- Added merge_view_mappings helper function
- Added parametrized pytest tests covering all 7 correctness properties
- Modified create_perforce_workspace_from_template to check P4_CLIENTS_ROOT_DIRECTORY
- Stream workspaces: lookup by stream path, reuse if found, register if new
- View workspaces: single shared workspace which adds new view paths as discovered/used
- Non-persistent behavior unchanged when env var is not set
- Added unit tests for all reuse flows and error handling
- Set Host='' in spec when P4_CLIENTS_ROOT_DIRECTORY is set
- Added tests for Host field behavior in both persistent and non-persistent modes

When creating perforce clients on workers we want to try to persist the client name(s) and views in order to minimize recreation of clients or redownloading of assets in cases where the workspace is persisted between jobs.

Write out a data file after a client is created/update with records of the clients created and the views used.  When initializing a perforce workspace, check for this file first and reuse clients where possible before trying to create a new one.

Better persistence and reuse of clients between jobs where possible.

All unit tests pass

Submitted perforce jobs to new and reused workers/clients.

Yes

Yes

No

----

### DIFF
--- a/src/deadline/unreal_perforce_utils/app.py
+++ b/src/deadline/unreal_perforce_utils/app.py
@@ -1,5 +1,7 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 
+import dataclasses
+import hashlib
 import os
 import json
 import pprint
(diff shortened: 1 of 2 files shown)


src/deadline/unreal_perforce_utils/app.py
```diff
@@ -14,6 +16,133 @@
...
+        except OSError as e:
+            logger.warning(
+                f"Could not read workspace info file at {path}: {e}. Starting with empty registry."
+            )
+            return WorkspaceInfoFile()
+
+        return WorkspaceInfoFile(
+            stream_workspaces=data.get("stream_workspaces", {}),
+            view_workspace=data.get("view_workspace"),
+        )
```
Cherie-Chen, line 63: NIT: While the risk of data corruption is low in the current implementation, adding validation here provides defense-in-depth and ensures forward compatibility if we modify the data layout in future iterations.

Here is what Kiro wrote for validation:
```
   # Validate types — fall back to empty on bad data
   stream_workspaces = data.get("stream_workspaces", {})
   view_workspace = data.get("view_workspace")

   if not isinstance(stream_workspaces, dict) or not all(
       isinstance(k, str) and isinstance(v, str) for k, v in stream_workspaces.items()
   ):
       logger.warning(f"Invalid stream_workspaces in {path}, starting with empty registry.")
       return WorkspaceInfoFile()

   if view_workspace is not None and not isinstance(view_workspace, str):
       logger.warning(f"Invalid view_workspace in {path}, starting with empty registry.")
       return WorkspaceInfoFile()

   return WorkspaceInfoFile(stream_workspaces=stream_workspaces, view_workspace=view_workspace)
```

Cherie-Chen (commented):
Code changes look good. I'll run manual tests before approving.

Future consideration (out of scope for this PR): We could add `created_at` or `last_used` timestamps to the `WorkspaceInfoFile` class to enable age-based cleanup of Perforce workspaces before Workers are terminated. However, we should defer implementing this until we have a concrete cleanup strategy in place.

baxeaz:
> Code changes look good. I'll run manual tests before approving.
> 
> Future consideration (out of scope for this PR): We could add `created_at` or `last_used` timestamps to the `WorkspaceInfoFile` class to enable age-based cleanup of Perforce workspaces before Workers are terminated. However, we should defer implementing this until we have a concrete cleanup strategy in place.

Hmm yeah we could maybe clean up some of them that way - the problem would still be workers/volumes/workspaces which were used within the time window, then the volume TTL'd out or was deleted and never got a chance to go through cleanup.

src/deadline/unreal_perforce_utils/app.py
baxeaz: Sure, can add that :)

Cherie-Chen (approved):
I have tested it in UE5.6 and it works great! Thanks!

leon-li-inspire (approved):

PR #274, opened 2026-03-31
feat: Notify customers if newer submitter is available.

karthikbekalp:
**DO NOT MERGE UNTIL https://github.com/aws-deadline/deadline-cloud/pull/1070 is released in Deadline Cloud**

Customers are not aware if they are working on the latest submitter or not. 

Created a new dialog to suggest customers to update their submitter. 

Customers hopefully have the latest submitter always or at least are notified of this. 

image

image

If customers want to deactivate the feature, they can do it in the settings. 

image

Yes

Also added a spec test. 

image

No as this dialog opens even before job submission. 

Yes added it into setup-submitter if customers want to deactivate the feature. 

No

----

### DIFF
--- a/src/unreal_plugin/Content/Python/update_check.py
+++ b/src/unreal_plugin/Content/Python/update_check.py
@@ -1,0 +1,151 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+"""
+Update checker for the Deadline Cloud for Unreal Engine plugin.
+
+Checks the GitHub releases API for a newer version and shows an Unreal
+Editor dialog when an update is available.  Respects the Deadline Cloud
+``settings.submitter_update_notification`` config toggle.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import os
+import socket
+import ssl
+import urllib.request
+import urllib.error
+import webbrowser
+
+import botocore
+import unreal
+
+from packaging.version import Version, InvalidVersion
+
+from deadline.client.config import config_file
+from deadline.unreal_submitter._version import version as _current_version
+
+logger = logging.getLogger(__name__)
+
+GITHUB_LATEST_RELEASE_URL = (
+    "https://api.github.com/repos/aws-deadline/deadline-cloud-for-unreal-engine/releases/latest"
+)
+RELEASES_PAGE_URL = "https://github.com/aws-deadline/deadline-cloud-for-unreal-engine/releases"
+_REQUEST_TIMEOUT_SECONDS = 5
+
+
+def _is_update_notification_enabled() -> bool:
+    """Check whether the user has opted in to update notifications."""
+    return config_file.str2bool(config_file.get_setting("settings.submitter_update_notification"))
+
+
+def _get_current_version() -> str:
+    """Return the currently installed plugin version string."""
+    return _current_version
+
+
+def _fetch_latest_version() -> str | None:
+    """Fetch the latest release tag from GitHub.
+
+    Returns:
+        The version string (e.g. ``"0.6.5"``) or ``None`` on failure.
+    """
+    # Pin to GitHub REST API v3 JSON format so the response shape stays stable.
+    req = urllib.request.Request(
+        GITHUB_LATEST_RELEASE_URL,
+        headers={"Accept": "application/vnd.github.v3+json"},
+    )
+
+    # Build a strict TLS context: enforce TLS 1.2+ and use the botocore CA
+    # bundle so certificate verification works even in embedded Python
+    # environments (e.g. Unreal) that may lack system root certificates.
+    ctx = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
+    ctx.minimum_version = ssl.TLSVersion.TLSv1_2
+    ctx.load_verify_locations(_get_botocore_ca_bundle())
+
+    try:
+        with urllib.request.urlopen(req, timeout=_REQUEST_TIMEOUT_SECONDS, context=ctx) as resp:
+            data = json.loads(resp.read().decode("utf-8"))
+            tag = data.get("tag_name", "")
+            return tag.lstrip("v") if tag else None
+    except urllib.error.URLError:
+        return None
+    except (socket.timeout, TimeoutError):
+        return None
+    except json.JSONDecodeError:
+        return None
+
+
+def _get_botocore_ca_bundle() -> str:
+    """Return the path to botocore's bundled CA certificate bundle."""
+    return os.path.join(os.path.dirname(botocore.__file__), "cacert.pem")
+
+
+def _is_update_available(current: str, latest: str) -> bool:
+    """Return True if *latest* is strictly newer than *current*."""
+    try:
+        return Version(latest) > Version(current)
+    except InvalidVersion:
+        return False
+
+
+def safe_check_and_show_update_dialog() -> bool:
+    """Check GitHub for a newer release and show an Unreal dialog if found.
+
+    Returns:
+        ``True`` if the user chose to open the download page (caller may
+        want to skip opening the submitter), ``False`` otherwise.
+    """
+    try:
+        return _check_and_show_update_dialog()
+    except Exception:
+        logger.debug("Update check failed -- skipping", exc_info=True)
+        return False
+
+
+def _check_and_show_update_dialog() -> bool:
+    """Internal implementation of the update check and dialog flow."""
+    if not _is_update_notification_enabled():
+        return False
+
+    current_version = _get_current_version()
+
+    latest_version = _fetch_latest_version()
+
+    if not latest_version:
+        return False
+
+    if not _is_update_available(current_version, latest_version):
+        return False
+
+    message = (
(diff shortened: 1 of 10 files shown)


leon-li-inspire (dismissed):

src/unreal_plugin/Content/Python/update_check.py
```diff
@@ -0,0 +1,151 @@
...
+
+    response = unreal.EditorDialog.show_message(
+        "New version available",
+        message,
+        unreal.AppMsgType.YES_NO,
+    )
+
+    if response == unreal.AppReturnType.YES:
+        try:
+            webbrowser.open(RELEASES_PAGE_URL)
```
joel-wong-aws, line 143: This will redirect users to the [release history page](https://github.com/aws-deadline/deadline-cloud-for-unreal-engine/releases), but how do they actually upgrade? I think we should be pointing to the actual docs that instruct them about _how_ to upgrade (is that [these](https://github.com/aws-deadline/deadline-cloud-for-unreal-engine/blob/mainline/docs/user_guide/setup-submitter.md#submitter-installation)?)

karthikbekalp: I had the same query, I discussed with @Cherie-Chen and then the suggestion was to use the release history page.

Cherie-Chen: I think we should have user review the release note and decide if they want to upgrade.

It's common for UE users to stick to one software version for the entire project. This can help them review the change set and decide if they want to upgrade

src/unreal_plugin/Content/Python/update_check.py
```diff
@@ -0,0 +1,151 @@
...
+
+    if not _is_update_available(current_version, latest_version):
+        return False
+
+    message = (
+        f"Version {latest_version} of Deadline Cloud for Unreal Engine "
+        f"submitter is now available.\n\n"
+        f"Current: {current_version}  ->  New: {latest_version}\n\n"
+        f"View release notes:\n{RELEASES_PAGE_URL}\n\n"
+        "Click 'Yes' to be redirected to the release page, or 'No' to dismiss."
```
joel-wong-aws: Can we add a message here about how to turn off the automatic checks via the GUI?

src/unreal_plugin/Content/Python/update_check.py
joel-wong-aws: This should link to https://aws-deadline.github.io/unreal-engine/setup-submitter/

test/deadline_submitter_for_unreal/unit/test_update_check.py
```diff
@@ -0,0 +1,180 @@
...
+    """Tests for _is_update_available()."""
+
+    def test_newer_version_available(self):
+        assert _is_update_available("0.5.0", "0.6.5") is True
+
+    def test_same_version(self):
+        assert _is_update_available("0.6.5", "0.6.5") is False
+
+    def test_older_version(self):
+        assert _is_update_available("1.0.0", "0.6.5") is False
```
joel-wong-aws: How about something like `0.10.0` and `0.6.5`? 

Can we make these into parameterized tests?

src/unreal_plugin/Source/UnrealDeadlineCloudService/Private/Tests/DeadlinePluginTest_UpdateDialog.spec.cpp
```diff
@@ -0,0 +1,177 @@
...
+		Settings->WorkStationConfiguration.General.ShowUpdateNotifications);
+
+	// Toggle on
+	Settings->WorkStationConfiguration.General.ShowUpdateNotifications = true;
+	TestTrue(
+		TEXT("ShowUpdateNotifications should be true after setting to true"),
+		Settings->WorkStationConfiguration.General.ShowUpdateNotifications);
+
+	// Restore
+	Settings->WorkStationConfiguration.General.ShowUpdateNotifications = bOriginal;
```
joel-wong-aws: I don't have too deep of an understanding of Unreal syntax, but to me this seems like setting and variable and then reading its value and asserting it is equal to what we just set it to. Am I missing something?

src/unreal_plugin/Source/UnrealDeadlineCloudService/Private/Tests/DeadlinePluginTest_UpdateDialog.spec.cpp
```diff
@@ -0,0 +1,177 @@
...
+{
+	const UDeadlineCloudDeveloperSettings* Settings = UDeadlineCloudDeveloperSettings::Get();
+	TestNotNull(TEXT("DeveloperSettings const singleton must exist"), Settings);
+	if (!Settings) return false;
+
+	// Just reading the value should not crash — the property must be well-formed.
+	const bool bValue = Settings->WorkStationConfiguration.General.ShowUpdateNotifications;
+	// We don't assert a specific value here because a previous test run may have
+	// changed it; we only verify the read succeeds without error.
+	TestTrue(TEXT("Reading ShowUpdateNotifications should succeed (value is bool)"), bValue || !bValue);
```
joel-wong-aws: Is this needed? Seems redundant

src/unreal_plugin/Source/UnrealDeadlineCloudService/Private/Tests/DeadlinePluginTest_UpdateDialog.spec.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+	Settings->WorkStationConfiguration.General.ShowUpdateNotifications = true;
+	TestTrue(
+		TEXT("Setting true should yield true"),
+		Settings->WorkStationConfiguration.General.ShowUpdateNotifications);
+
+	Settings->WorkStationConfiguration.General.ShowUpdateNotifications = false;
+	TestFalse(
+		TEXT("Setting false should yield false"),
+		Settings->WorkStationConfiguration.General.ShowUpdateNotifications);
```
joel-wong-aws, line 99: Similar note as above about this simply asserting `true == true` - this doesn't seem to actually set a string value of `true` as the test name indicates

src/unreal_plugin/Content/Python/update_check.py
karthikbekalp: I'd pushback on this as it would make it a bunch of text that customers might not read because of its length. If this becomes a problem, we can always add more info.

src/unreal_plugin/Content/Python/update_check.py
karthikbekalp: Wait, didn't we confirm that we need to use release page here as we still want customers to review the release note and decide if they want to upgrade.

test/deadline_submitter_for_unreal/unit/test_update_check.py
karthikbekalp: Sure, I'll add it.

src/unreal_plugin/Source/UnrealDeadlineCloudService/Private/Tests/DeadlinePluginTest_UpdateDialog.spec.cpp
karthikbekalp: Hmm, my bad. I used the skill to generate the tests but didn't review them. Will work on fixing it and adding useful tests.

src/unreal_plugin/Source/UnrealDeadlineCloudService/Private/Tests/DeadlinePluginTest_UpdateDialog.spec.cpp
karthikbekalp: Hmm, my bad. I used the skill to generate the tests but didn't review them. Will work on fixing it and adding useful tests.

src/unreal_plugin/Source/UnrealDeadlineCloudService/Private/Tests/DeadlinePluginTest_UpdateDialog.spec.cpp
karthikbekalp: Hmm, my bad. I used the skill to generate the tests but didn't review them. Will work on fixing it and adding useful tests.

test/deadline_submitter_for_unreal/unit/test_update_check.py
karthikbekalp: Updated.

src/unreal_plugin/Content/Python/update_check.py
karthikbekalp: Added the instructions to turn off automatic checks.

src/unreal_plugin/Content/Python/update_check.py
karthikbekalp: Discussed offline, once customers review the release notes the next dialog would have the instructions to update the submitter.

joel-wong-aws (dismissed):

Cherie-Chen (dismissed):

Cherie-Chen (approved):

leon-li-inspire (approved):

PR #295, opened 2026-05-07
feat: add e2e test options for external infrastructure and conda chan…

baxeaz:

The end-to-end test suite always created its own farm/queue/fleet and used
the configured CondaChannels from the render job template. That made it
hard to run e2e tests against pre-existing infrastructure or against an alternate conda channel.
Separately, three issues surfaced while iterating on this branch:

- `max_retries_per_task` defaulted to 50, which is far too high for a
  shared default — failing tasks would be retried for hours before the
  job surfaced the failure.
- `DeadlineCloudDetailsWidgetsHelper.cpp` referenced
  `SHorizontalBox::Slot().MinWidth(...)`, which is not compatible with older
  versions of Unreal.
- When only doing an --install (not the --worker option) we were installing to the global python environment which may require admin rights and should not be required.

- Added pytest options to the e2e suite: `--farm-id`, `--queue-id`,
  `--no-cancel`, and `--conda-channel`. `reusable_farm_id` and
  `reusable_queue_id` now prefer the CLI override before falling back to
  deadline config / queue creation, and `test_create_job` honors
  `--no-cancel` so the job can flow into downstream worker-agent tests.
- Added a `DEADLINE_CONDA_CHANNELS` env override in `job_library.py` that
  rewrites the `CondaChannels` parameter at submission time. The e2e
  `patch_render_template` fixture sets the env var when
  `--conda-channel` is passed; no template files are mutated on disk.
- Lowered `JobSharedSettings.max_retries_per_task` default from 50 to 2.
- Removed the `MinWidth(200.f)` call and converted two local lambdas to
  `TFunction` (and dropped an unused `FPropertyHandleItemAddResult`
  binding) so the plugin compiles on pre c++ 20 compilers
- Only perform the global pip install (Where we install our adaptor and python utilities) when the worker option is supplied

- E2E tests can be pointed at existing farms/queues and an alternate conda
  channel without editing config or template files. Default behavior
  (create-and-reuse) is preserved when the new options are omitted.
- Submissions inherit a saner retry default; users who want the previous
  value can override `max_retries_per_task` per job.
- The submitter plugin once again builds on older Unreal versions

- Ran the existing e2e suite locally against a pre-existing farm and queue
  using `--farm-id`, `--queue-id`, and `--no-cancel`, and confirmed the
  job reached READY without being canceled.
- Ran the same suite with `--conda-channel <test-channel>` and verified
  the submitted job's `CondaChannels` parameter reflected the override.
- Built the unreal plugin against UE 5.4 and 5.5 to confirm the
  `MinWidth`/lambda changes compile.
- Yes — unit tests pass.

Yes — submitted and rendered a test job through the patched submitter
against a configured farm/queue.

Yes

No. The new pytest options default to the prior behavior, the
`DEADLINE_CONDA_CHANNELS` env var only takes effect when explicitly set,
and lowering `max_retries_per_task` only changes the *default* — any
template or job that sets the value explicitly is unaffected.

----

### DIFF
--- a/test/end_to_end/conftest.py
+++ b/test/end_to_end/conftest.py
@@ -220,6 +220,30 @@ def pytest_addoption(parser) -> None:
         default=False,
         help="Clean up resources (queues, fleets, associations) after tests",
     )
+    parser.addoption(
+        "--farm-id",
+        action="store",
+        default=None,
+        help="Use a specific farm ID instead of reading from deadline config",
+    )
+    parser.addoption(
+        "--queue-id",
+        action="store",
+        default=None,
+        help="Use a specific queue ID instead of creating/reusing a test queue",
+    )
+    parser.addoption(
+        "--no-cancel",
+        action="store_true",
+        default=False,
+        help="Don't cancel the job after it reaches READY state",
+    )
+    parser.addoption(
+        "--conda-channel",
+        action="store",
+        default=None,
+        help="Override the CondaChannels default in the render job template",
+    )
 
 
 def get_source_root() -> str:
@@ -626,13 +650,14 @@ def extract_job_info_from_test_output(
 
 
 @pytest.fixture
-def run_unreal_test(request, reusable_queue_fleet_association) -> Callable:
+def run_unreal_test(request, reusable_farm_id, reusable_queue_id) -> Callable:
     """
     Fixture that provides a function to run Unreal Engine automation tests.
 
     Args:
         request: The pytest request object
-        reusable_queue_fleet_association: Fixture providing farm, queue, and fleet IDs
+        reusable_farm_id: The farm ID
+        reusable_queue_id: The queue ID
 
     Returns:
         A callable function that runs Unreal Engine automation tests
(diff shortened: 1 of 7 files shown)


Cherie-Chen (dismissed):

test/end_to_end/conftest.py
```diff
@@ -657,13 +688,13 @@ def _run_unreal_test(
         if deadlineargs is None:
             deadlineargs = "-NoLoadingScreen -FixedSeed -log -Unattended -MRQInstance -deterministicaudio -audiomixer"
 
-        reusable_farm_id, reusable_queue_id, reusable_fleet_id = reusable_queue_fleet_association
+        reusable_farm_id_val, reusable_queue_id_val = reusable_farm_id, reusable_queue_id
```
leon-li-inspire: What's the reason this variable rebinding ?

test/end_to_end/conftest.py
```diff
@@ -754,39 +785,51 @@ def build_plugin(request) -> None:
...
 
-    # Run the script and capture the output
-    result = subprocess.run(build_args, text=True)
-    assert result.returncode == 0
+    Sets DEADLINE_CONDA_CHANNELS env var which the plugin reads at submission time.
+    Only sets when --conda-channel is explicitly passed. No template files are modified.
+    """
+    conda_channel = request.config.getoption("--conda-channel")
+    if conda_channel:
+        os.environ["DEADLINE_CONDA_CHANNELS"] = conda_channel
```
leon-li-inspire: Do we need to restore DEADLINE_CONDA_CHANNELS after the test ?

test/end_to_end/conftest.py
```diff
@@ -220,6 +220,36 @@ def pytest_addoption(parser) -> None:
...
+        help="Don't cancel the job after it reaches READY state",
+    )
+    parser.addoption(
+        "--conda-channel",
+        action="store",
+        default=None,
+        help="Override the CondaChannels default in the render job template",
+    )
+    parser.addoption(
+        "--conda-packages",
```
leon-li-inspire: Are we using this option any place ?

test/end_to_end/conftest.py
baxeaz: Good idea, fixed!

test/end_to_end/conftest.py
baxeaz: Unnecessary, removed!

test/end_to_end/conftest.py
baxeaz: Nope, removed!

leon-li-inspire (approved):

Cherie-Chen (approved):

Issue #296, opened 2026-05-09
Bug: unreal-engine-openjd 0.6.8 + UE 5.7: EditorAssetLibrary.load_asset() returns None for any LevelSequence on render workers


akre54:
### Describe the bug

After migrating a UE 5.6 project to UE 5.7 and re-packaging the conda env via the
deadline-cloud channel (`unrealengine=5.7 unrealengine-openjd`), the deadline-cloud-for-unreal
adaptor (v0.6.8) fails on every render with:

```
LogEditorAssetSubsystem: Error: LoadAsset failed:
LogPython: Error: Render Executor: Error: Level Sequence not loaded.
Check if the sequence exists and is valid
```

followed by:

```
File ".../unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py",
line 319, in get_frame_range
level_sequence.get_playback_range().get_start_frame()
AttributeError: 'NoneType' object has no attribute 'get_playback_range'
```

The `EditorAssetLibrary.load_asset(level_sequence_path)` call at
[step_handlers/unreal_render_step_handler.py L72-76](https://github.com/aws-deadline/deadline-cloud-for-unreal-engine/blob/release/src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py#L72-L76)
returns `None` for **any** level sequence on the rendering worker, even though:
- Plugins all load successfully
- Asset registry has been loaded (105 MiB cache restored, `AssetRegistry took 0.0051s to start up`,
  `AssetRegistryGather` completed before `run_script`)
- The .uasset file is present on disk at the expected path
- The same project loads & renders fine when opened in UE 5.7 Editor on a developer workstation
  (Mac), where `unreal.load_asset(...)` works without issue
- The same workflow worked under UE 5.6 with this adaptor reproduces with multiple distinct sequence assets in the same project.

### Expected Behaviour

Expected behavior: `unreal-engine-openjd 0.6.8` should render any LevelSequence that loads in UE 5.7 Editor, the same way it did under UE 5.6. Specifically, `EditorAssetLibrary.load_asset(level_sequence_path)` should return the loaded sequence (not `None`) when the asset registry is populated and the .uasset is on disk, so `get_frame_range()` can read `playback_range` and the render proceeds. As a defense in depth, the `None` case in `get_frame_range()` should log + return a usable range rather than crash with `AttributeError`.

### Current Behaviour

```
[...]

PR #298, opened 2026-05-12
fix: guard adaptor against None LevelSequence + force custom playback range in chunked renders

akre54:
## Summary

> **Note on scope** — an earlier revision of this PR proposed reordering the `LevelSequence` loader chain (`unreal.load_asset` first, `EditorAssetLibrary.load_asset` as fallback) under the hypothesis that the latter silently returns `None` under UE 5.7 + `-RenderOffscreen`. An empirical sweep across UE 5.4 / 5.5 / 5.6 / 5.7 against a freshly-created `LevelSequence` did **not** reproduce that failure mode — both loaders worked on every version. Without a verifiable repro the reorder was speculative, so I've dropped it. This PR is now narrowed to two defensive changes whose evidence is unambiguous, both relevant to #296.

### Change 1 — `get_frame_range` is None-guarded

`UnrealRenderStepHandler.get_frame_range` previously called `level_sequence.get_playback_range()` unconditionally. If the upstream loader returned `None` for any reason, the adaptor would crash with `AttributeError: 'NoneType' object has no attribute 'get_playback_range'` — which is the exact trace we have observed in production.

New behaviour: when `level_sequence is None` and `use_custom_playback_range` is false:
- If `output_settings.custom_end_frame > custom_start_frame`, use that MRQ custom range as a fallback and log a warning.
- Otherwise, return `(None, None)` so the caller can surface a clear error instead of crashing on attribute access. (The chunk path in `run_script` checks for this and returns `False`.)

This change does not attempt to claim a cause for why `level_sequence` might be `None`. It just refuses to crash on the resulting attribute access.

### Change 2 — force `use_custom_playback_range = True` in the `frames_per_task` chunk path

When `frames_per_task` + `chunk_id` are set, the chunk path computes a per-chunk `custom_start_frame` and `custom_end_frame` on `output_settings`. **But it never set `use_custom_playback_range = True`.** MRQ ignores the custom frame values when that flag is false and falls back to the playback range baked into the MRQ preset.

In a 40-chunk render with `frames_per_task` set, every task therefore rendered the full sequence rather than its assigned window — wasting ~40× the render time and producing last-writer-wins output frames.

This is provable by reading MRQ source (`MoviePipelineOutputSetting.cpp`) — `use_custom_playback_range` gates the use of `custom_start_frame` / `custom_end_frame` in `GetEffectivePlaybackRange`. Setting the bounds without flipping the flag is a no-op.

The fix is a one-liner: set the flag after computing the chunk window. This PR also adds a None-guard so that if the level sequence didn't load, the chunk window is still applied via `output_settings.use_custom_playback_range` (since `level_sequence.set_playback_start/end` are no-ops on `None`).

## Test plan

- [x] **Unit tests added** (`test_unreal_render_step_handler_none_guard.py`): four cases pinning `get_frame_range` behaviour:
  - None sequence + non-empty custom range → falls back to custom range, warns
  - None sequence + empty custom range → `(None, None)`, errors
  - `use_custom_playback_range = True` + None sequence → short-circuits regardless of None
  - Non-None sequence → uses `sequence.get_playback_range()` as before (regression guard)
- [x] **Existing tests pass**: 29 step-handler unit tests green (`pytest test/deadline_adaptor_for_unreal --no-cov`).
- [x] **Lint clean**: `ruff check` + `black --check` on the changed files.
- [x] **Empirical sweep against UE 5.4 / 5.5 / 5.6 / 5.7** (Mac, headless `UnrealEditor-Cmd -unattended -nullrhi`, freshly-created `LevelSequence`): both `unreal.load_asset` and `EditorAssetLibrary.load_asset` worked on every version. This is what convinced me to drop the loader reorder.

## What this PR is and is not

- **Is**: a defensive hardening against `NoneType` attribute access (the actual production crash trace) and a fix for an MRQ `use_custom_playback_range` gating bug (independently verifiable from MRQ source). Both are scoped to the chunked-render path that the OpenJD adaptor uses.
- **Is not**: a claim about what causes `level_sequence` to be `None`, nor a fix for the speculative `EditorAssetLibrary.load_asset` UE 5.7 regression that the original #296 report alleged. If that regression exists, it's not the one this PR fixes — and I no longer have evidence that it exists at all.

Refs #296. (Earlier conversation on #296 referenced a loader-reorder fix; the comment on the issue has been updated with the narrower framing.)

Signed-off-by: Adam Krebs <931368+akre54@users.noreply.github.com>

### DIFF
--- a/src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
+++ b/src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
@@ -75,15 +75,35 @@ if unreal:
                         )
                     )
                     if level_sequence is None:
-                        logger.error(
-                            "Render Executor: Error: Level Sequence not loaded. Check if the sequence "
-                            "exists and is valid"
+                        # Defensive fallback: if the LevelSequence can't be loaded
+                        # (we have seen this happen in production for reasons that
+                        # are not always reproducible), use the MRQ output_settings
+                        # custom range if it is non-empty. This avoids crashing the
+                        # render with an `AttributeError: 'NoneType' object has no
+                        # attribute 'get_playback_end'` when something upstream
+                        # caused the loader to return None.
+                        if output_settings.custom_end_frame > output_settings.custom_start_frame:
+                            logger.warning(
+                                "Render Executor: Level Sequence not loaded; falling back to "
+                                f"output_settings custom range "
+                                f"[{output_settings.custom_start_frame}, "
+                                f"{output_settings.custom_end_frame}]"
+                            )
+                            self.totalFrameRange += (
+                                output_settings.custom_end_frame
+                                - output_settings.custom_start_frame
+                            )
+                        else:
+                            logger.error(
+                                "Render Executor: Error: Level Sequence not loaded and "
+                                "output_settings has no custom range. Check if the sequence "
+                                "exists and is valid."
+                            )
+                            return
+                    else:
+                        self.totalFrameRange += (
+                            level_sequence.get_playback_end() - level_sequence.get_playback_start()
                         )
-                        return
-
-                    self.totalFrameRange += (
-                        level_sequence.get_playback_end() - level_sequence.get_playback_start()
-                    )
 
                 if self.totalFrameRange == 0:
                     logger.error(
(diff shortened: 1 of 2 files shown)


Cherie-Chen:
Thanks @akre54 — really appreciate the thoroughness here. The code looks good to me. I'll run our internal end-to-end validation pass across the supported UE versions and worker configurations before formally approving. 

Thanks again for the contribution and the quality of the write-up throughout #296 and this PR!

Cherie-Chen (approved):

baxeaz:
Echoing what @Cherie-Chen said, thanks so much for the deep dive and PR @akre54 !

baxeaz (approved):

PR #306, opened 2026-06-02
test: add automated integration tests for bundle generation

kavmur:
The Unreal Engine Deadline Cloud plugin has no automated integration tests in CI. Currently, only pure Python unit tests (mocked, no UE required) run automatically on PRs/merges. The end-to-end tests (test/end_to_end/) require manual execution by an engineer before approving the release gate.

Added local integration tests that:
- Launch UE 5.7 headless 
- Generate a job bundle 
- Validate the YAML 

Every PR and mainline push will validate that UE can generate valid OpenJD job bundles (once CDK infrastructure is deployed — separate CR)

Validated the tests on dev EC2 instance and also on CodeBuild gamma.

Yes — 467 passed, 2 skipped, 0 failures.

yes

NA

Only adds tests.

----

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -1,0 +1,45 @@
+name: Unreal Engine Integration Tests
+
+on:
+  push:
+    branches:
+      - mainline
+  workflow_dispatch:
+    inputs:
+      ref_type:
+        description: 'Reference type to test'
+        required: true
+        default: 'branch'
+        type: choice
+        options:
+          - branch
+          - tag
+      branch:
+        description: 'Branch to test (e.g., mainline, feature-branch)'
+        required: false
+        default: 'mainline'
+        type: string
+      tag:
+        description: 'Tag to test (only used when ref_type is "tag")'
+        required: false
+        type: string
+
+jobs:
+  IntegrationTests:
+    name: ${{ matrix.name }} Integration Tests
+    permissions:
+      id-token: write
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: windows
+            name: Windows
+    uses: aws-deadline/.github/.github/workflows/reusable_integration_test.yml@mainline
+    secrets: inherit
+    with:
+      repository: ${{ github.event.repository.name }}
+      branch: ${{ github.event.inputs.ref_type == 'branch' && github.event.inputs.branch || 'mainline' }}
+      tag: ${{ github.event.inputs.ref_type == 'tag' && github.event.inputs.tag || '' }}
+      os: ${{ matrix.os }}
--- a/pipeline/generate_bundle.py
+++ b/pipeline/generate_bundle.py
@@ -1,0 +1,97 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+"""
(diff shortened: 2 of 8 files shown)


.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,48 @@
+name: Unreal Engine Integration Tests
+
+on:
+  push:
+    branches:
+      - mainline
+  pull_request:
+    branches:
+      - mainline
```
crowecawcaw: This won't work because `pull_request` events don't trigger workflows with secrets, and secrets are needed for integ tests. We also don't want untrusted PRs to get secrets. Better to just remove it - we'll run integ tests once a PR is approved and merged. Not as convenient, but safer.

crowecawcaw:
Which parts of the integration does this cover and which are not covered? Seems like it's testing a piece of the submitter, but since it runs a python script through Unreal, some submitter parts aren't in scope. Might be ok depending on what we aim to test.

.github/workflows/integration_tests.yml
kavmur: makes sense, removed pull_request trigger. Thanks!

kavmur:
> Which parts of the integration does this cover and which are not covered? Seems like it's testing a piece of the submitter, but since it runs a python script through Unreal, some submitter parts aren't in scope. Might be ok depending on what we aim to test.

Right currently the test only checks if UE launches headlessly with the Deadline Cloud plugin loaded
and produces valid OpenJD YAML (template, parameters, asset references). I will be adding more tests once I can verify the infra is setup and is working.

pipeline/setup-runner.py
```diff
@@ -0,0 +1,270 @@
...
+    print("Installing Visual Studio Build Tools 2022...")
+    installer_path = Path("C:/Temp/vs_buildtools.exe")
+    installer_path.parent.mkdir(parents=True, exist_ok=True)
+
+    run(
+        [
+            "powershell",
+            "-Command",
+            "Invoke-WebRequest -Uri 'https://aka.ms/vs/17/release/vs_buildtools.exe'"
+            f" -OutFile '{installer_path}' -UseBasicParsing",
```
leon-li-inspire, line 154: Do we need to checksum for this download file too ?

pipeline/generate_bundle.py
```diff
@@ -0,0 +1,97 @@
...
+    except Exception as e:
+        unreal.log_error(f"generate_bundle: Failed - {e}")
+        unreal.log_error(traceback.format_exc())
+        # Write error to marker so test harness can report the failure.
+        try:
+            os.makedirs(os.path.dirname(marker_file), exist_ok=True)
+            with open(marker_file, "w", encoding="utf-8") as f:
+                f.write(f"ERROR: {e}")
+        except Exception:
+            pass  # Best-effort: marker write failed, UE will still exit below
```
leon-li-inspire: Could we have a log here ?

pipeline/setup-runner.py
kavmur: Reason I used the web based downloads is because the URL always points to the latest stable release and a pinned checksum would break on any updates. For other S3-hosted artifacts (UE zip, 7-Zip) we do verify checksums since those are versioned.

leon-li-inspire (approved):

crowecawcaw (approved):

PR #307, opened 2026-06-05
docs: remove migrated Unreal Engine user guide

crowecawcaw:
### What / Why
The Unreal Engine user guide was migrated into the AWS Deadline Cloud service documentation, so the copy in this repo is now duplicated content that can drift.

### How
- Remove `docs/user_guide/` and the MkDocs config that published it.
- Remove `docs/telemetry.md` (the generic opt-out is documented centrally in the service docs).
- Repoint README links to the user guide (`userguide/epic-unreal-engine.html`) and telemetry opt-out (`userguide/opt-out.html`).

Non-migrated developer docs are retained. Docs-only change — no code or behavior affected.

### Breaking change?
No.


Cherie-Chen:
The current user guide is organized across multiple pages with a focused goal per page. The new service doc puts everything into one long page, which makes it harder to navigate.

I'd prefer we keep the user guide as-is until the new doc is better organized. I'm happy to take a look at improving it.

PR #312, opened 2026-06-11
test: add e2e integration tests for adaptor, job creation, and worker agent

kavmur:

`deadline-cloud-for-unreal-engine` had no automated end-to-end test coverage for the adaptor lifecycle, job submission, or worker agent rendering. The only CI test was bundle generation (PR #306), which validates YAML structure but not complete plugin and submission workflow end to end. 

**Tests added (4 tests across 2 CI workflows):**
- `test_bundle_generation` — validates job bundle YAML + OpenJD schema (already shipped in #306)
- `test_openjd_run` — validates adaptor daemon lifecycle (start → named pipe → script exec → stop) via `openjd run`
- `test_create_job` — smoke test: UE automation submits a job, verifies READY state
- `test_worker_agent_project_plugins` — full round-trip: submit → local worker picks up → adaptor renders → SUCCEEDED, validates only enabled content plugins loaded

**Source fixes :**

- `remote_executor.py`: switched to `MoviePipelinePythonHostExecutor` and override `execute_delayed()`.  This is required because python UClasses  may not be fully registered when MRQ first calls Execute() (specially in headless mode). [MoviePipelinePythonHostExecutor](https://dev.epicgames.com/documentation/en-us/unreal-engine/python-api/class/MoviePipelinePythonHostExecutor) is "designed to host an executor implemented in Python".
- `common.py`: added `os.path.isabs()` fallback checks. Original `convert_relative_path_to_full` behavior preserved; fallbacks only trigger in headless mode.

No customer-facing impact in interactive mode, and headless mode works!

- Full suite passing in gamma runner.

Yes. The `test_worker_agent_project_plugins` test submits a real job via UE automation, a local worker agent processes it with the UE adaptor, and the job completes with SUCCEEDED status. The `test_create_job` test also submits successfully and verifies READY state.

No documentation updates required. Test infrastructure is internal to the CI pipeline.

No. The `remote_executor.py` base class change maintains the same external behavior — `MoviePipelinePythonHostExecutor` inherits from `MoviePipelineExecutorBase`, so all C++ code that references the class via `TSubclassOf<UMoviePipelineExecutorBase>` continues to work. The `common.py` changes are purely additive fallbacks.

---

_By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice._

### DIFF
--- a/src/deadline/unreal_submitter/common.py
+++ b/src/deadline/unreal_submitter/common.py
@@ -279,6 +279,8 @@ def create_deadline_cloud_temp_file(file_prefix: str, file_data: Any, file_ext: 
         "UnrealDeadlineCloudService",
         file_prefix,
     )
+    if not os.path.isabs(destination_dir):
+        destination_dir = os.path.abspath(destination_dir)
     os.makedirs(destination_dir, exist_ok=True)
 
     temp_file = unreal.Paths.create_temp_filename(
@@ -292,8 +294,10 @@ def create_deadline_cloud_temp_file(file_prefix: str, file_data: Any, file_ext: 
         else:
             f.write(file_data)
 
-    temp_file = unreal.Paths.convert_relative_path_to_full(temp_file).replace("\\", "/")
-
+    temp_file = unreal.Paths.convert_relative_path_to_full(temp_file)
+    if not os.path.isabs(temp_file):
+        temp_file = os.path.abspath(temp_file)
+    temp_file = temp_file.replace("\\", "/")
     return temp_file
 
 
--- a/src/unreal_plugin/Content/Python/remote_executor.py
+++ b/src/unreal_plugin/Content/Python/remote_executor.py
@@ -10,13 +10,20 @@ logger = get_logger()
 
 
 @unreal.uclass()
-class MoviePipelineDeadlineCloudRemoteExecutor(unreal.MoviePipelineExecutorBase):
-    # The queue we are working on, null if no queue has been provided.
-    pipeline_queue = unreal.uproperty(unreal.MoviePipelineQueue)
+class MoviePipelineDeadlineCloudRemoteExecutor(unreal.MoviePipelinePythonHostExecutor):
+    """
+    Deadline Cloud remote executor for Movie Render Queue.
+
+    Inherits from MoviePipelinePythonHostExecutor (not MoviePipelineExecutorBase directly)
+    because Python-defined UClasses aren't available when the executor is first initialized.
+    The host executor's C++ code handles Execute() dispatch and forwards to execute_delayed()
+    which Python can reliably override.
+    """
+
     job_ids = unreal.uproperty(unreal.Array(str))
 
     @unreal.ufunction(override=True)
-    def execute(self, pipeline_queue):
+    def execute_delayed(self, pipeline_queue):
         logger.info(f"Asked to execute Queue: {pipeline_queue}")
         logger.info(f"Queue has {len(pipeline_queue.get_jobs())} jobs")
 
@@ -32,7 +39,7 @@ class MoviePipelineDeadlineCloudRemoteExecutor(unreal.MoviePipelineExecutorBase)
 
         self.pipeline_queue = pipeline_queue
 
-        unreal_submitter = UnrealMrqJobSubmitter()
+        unreal_submitter = UnrealMrqJobSubmitter(silent_mode=unreal.SystemLibrary.is_unattended())
 
         for job in self.pipeline_queue.get_jobs():
             logger.info(f"Submitting Job `{job.job_name}` to Deadline Cloud...")
@@ -42,12 +49,6 @@ class MoviePipelineDeadlineCloudRemoteExecutor(unreal.MoviePipelineExecutorBase)
 
     @unreal.ufunction(override=True)
     def is_rendering(self):
-        # Because we forward unfinished jobs onto another service when the
-        # button is pressed, they can always submit what is in the queue and
-        # there's no need to block the queue.
-        # A MoviePipelineExecutor implementation must override this. If you
-        # override a ufunction from a base class you don't specify the return
-        # type or parameter types.
         return False
 
     def check_dirty_packages(self) -> bool:
--- a/src/unreal_plugin/Source/UnrealDeadlineCloudService/Private/Tests/Integration/DeadlinePluginTest_CreateJobTest.cpp
+++ b/src/unreal_plugin/Source/UnrealDeadlineCloudService/Private/Tests/Integration/DeadlinePluginTest_CreateJobTest.cpp
@@ -90,11 +90,9 @@ public:
             m_renderStarted = true;
         }
 
-        if (m_jobCreationFound && m_dialogConfirmationFound)
+        if (m_jobCreationFound)
         {
-            UE_LOG(LogCreateJobTest, Display, TEXT("Both conditions met, marking test as successful"));
             m_testInstance->TestTrue("Job creation succeeded", true);
-
             return true;
         }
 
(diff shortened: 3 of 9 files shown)


src/deadline/unreal_adaptor/UnrealClient/step_handlers/__init__.py
```diff
@@ -3,15 +3,19 @@
...
 
 from .base_step_handler import BaseStepHandler
-from .unreal_render_step_handler import UnrealRenderStepHandler
 from .unreal_custom_step_handler import UnrealCustomStepHandler
 
+try:
+    from .unreal_render_step_handler import UnrealRenderStepHandler
+except (ImportError, AttributeError):
+    UnrealRenderStepHandler = None  # type: ignore[assignment,misc]
+
```
crowecawcaw: Why is this fallback here? Is this intentional?

src/deadline/unreal_adaptor/UnrealClient/unreal_client.py
```diff
@@ -12,6 +12,19 @@
         if p not in sys.path:
             sys.path.insert(0, p.replace("\\", "/"))
 
+# pywin32 native extensions need their DLL directory registered for the current process.
+# The adaptor adds pywin32 paths to PYTHONPATH, but os.add_dll_directory must be called here.
+if sys.platform == "win32":
```
crowecawcaw: There are a few of these. Unreal is only on Windows right? Do we have win32 checks elsewhere in this codebase?

src/deadline/unreal_submitter/common.py
```diff
@@ -292,8 +295,7 @@ def create_deadline_cloud_temp_file(file_prefix: str, file_data: Any, file_ext:
         else:
             f.write(file_data)
 
-    temp_file = unreal.Paths.convert_relative_path_to_full(temp_file).replace("\\", "/")
-
+    temp_file = temp_file.replace("\\", "/")
```
crowecawcaw, line 299: Is this used in other places? Seems like it was intentional and could break other tests

src/unreal_plugin/Content/Python/init_unreal.py
```diff
@@ -203,6 +203,22 @@ def __init__(
...
+        _win32_dir = os.path.join(_libraries_path, "win32")
+        _win32_lib = os.path.join(_libraries_path, "win32", "lib")
+        _pywin32_sys32 = os.path.join(_libraries_path, "pywin32_system32")
+        if os.path.isdir(_win32_dir) and _win32_dir not in sys.path:
+            sys.path.insert(0, _win32_dir)
+        if os.path.isdir(_win32_lib) and _win32_lib not in sys.path:
+            sys.path.insert(0, _win32_lib)
+        if os.path.isdir(_pywin32_sys32):
+            os.environ["PATH"] = _pywin32_sys32 + os.pathsep + os.environ.get("PATH", "")
+            os.add_dll_directory(_pywin32_sys32)  # type: ignore[attr-defined]
```
crowecawcaw: There are a lot of changes to the submitter and plugin itself, not just adding tests. Do we need them all? Ideally we'd add tests without touching the things being tested. Might not be possible, but there's a lot in this PR.

src/deadline/unreal_adaptor/UnrealClient/step_handlers/__init__.py
kavmur: yes intentional, when the adaptor runs in headless/CI mode MoviePipelinePIEExecutor will not exist as some plugins are not loaded in headless mode.

src/deadline/unreal_adaptor/UnrealClient/unreal_client.py
kavmur: correct, unreal plugin is windows only (we could extend support to linux/mac in future).  
The main reason for this check is our unit tests would fail on linux/mac runners without these checks. 

But it makes sense to move this out of the adapter source, as this is not needed in prod (where unreal is installed via conda). I can move this out of the source.

src/deadline/unreal_submitter/common.py
kavmur: From my tests `convert_relative_path_to_full` is actually a no-op in interactive mode as project_saved_dir() already returns full absolute path. 
In headless mode (our test), project_saved_dir() returns relative path and this fails as current working directory may be different. 
If its safer, I can keep convert_relative_path_to_full as a fallback but add os.path.abspath only when the result is still relative

src/unreal_plugin/Content/Python/init_unreal.py
kavmur: Makes sense, we dont need them all. For the changes in adaptor, I can move them out of the source as they are only required for our test worker agent. But for submitter and plugin changes some of them are more fixes to make it work in headless mode. 
I have added in PR description on the src changes. And the changes are only additive and shouldn't break current behavior which is why most are done as part of fallback.

src/deadline/unreal_submitter/common.py
kavmur: Revised to keep convert_relative_path_to_full and only call os.path.abspath as a fallback only if the paths are still relative.

src/deadline/unreal_adaptor/UnrealClient/unreal_client.py
kavmur: done

src/unreal_plugin/Content/Python/remote_executor.py
```diff
@@ -32,7 +39,7 @@ def execute(self, pipeline_queue):
 
         self.pipeline_queue = pipeline_queue
```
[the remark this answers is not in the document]
kavmur: Based on [this](https://dev.epicgames.com/documentation/en-us/unreal-engine/python-api/class/MoviePipelinePythonHostExecutor), pipeline_queue is a read-write property, setting `self.pipeline_queue = pipeline_queue` writes to the inherited property

src/unreal_plugin/Source/UnrealDeadlineCloudService/Private/Tests/Integration/DeadlinePluginTest_CreateJobTest.cpp
```diff
@@ -90,11 +90,9 @@ class WaitForJobCreationLogCommand : public IAutomationLatentCommand, public FOu
             m_renderStarted = true;
         }
 
-        if (m_jobCreationFound && m_dialogConfirmationFound)
+        if (m_jobCreationFound)
```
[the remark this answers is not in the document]
kavmur: This is intentional, as this test would now run exclusively in headless/CI mode (-unattended) where silent_mode=True suppresses all dialogs. We can add this as part of UI test in the future if required.

karthikbekalp (dismissed):

src/unreal_plugin/Content/Python/remote_executor.py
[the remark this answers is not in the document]
Cherie-Chen: The link in "this" is broken. Could you please find it again?

test/end_to_end/conftest.py
```diff
@@ -1632,71 +1639,80 @@ def deadline_worker_agent(
...
+    # Environment variables for the worker agent
     env = os.environ.copy()
     env["PYTHONIOENCODING"] = "utf-8"
-    env["TERM"] = "dumb"  # Disable terminal features
-    env["NO_COLOR"] = "1"  # Disable color output
+    env["TERM"] = "dumb"
+    env["NO_COLOR"] = "1"
+
+    # Add UE binaries to PATH so the adaptor can find UnrealEditor-Cmd
+    ue_bin_dir = os.path.join(find_engine_root(None), "Engine", "Binaries", "Win64")
```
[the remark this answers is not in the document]
Cherie-Chen: This a real problem. User can have multiple UE versions installed. Please address this

src/unreal_plugin/Content/Python/remote_executor.py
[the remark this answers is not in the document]
kavmur: https://dev.epicgames.com/documentation/en-us/unreal-engine/python-api/class/MoviePipelinePythonHostExecutor?application_version=4.27

test/end_to_end/conftest.py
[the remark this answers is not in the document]
kavmur: we are doing this only for the e2e test right? where we would have just one version?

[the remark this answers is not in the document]
Cherie-Chen: Discussed offline. We will bias-for-action here. This won't be an issue testing in CI/CD testing because each environment will have only 1 Ue version installed, but can be an issue with dev machine. I will confirm the behavior and cut an issue if that's the case.

Cherie-Chen (approved):

test/integ/test_openjd_run.py
```diff
@@ -0,0 +1,141 @@
...
+    logger.info(f"Running openjd run: {cmd}")
+    result = subprocess.run(cmd, capture_output=True, text=True, timeout=300)
+
+    logger.info(f"Return code: {result.returncode}")
+    logger.info(f"STDOUT:\n{result.stdout}")
+    if result.stderr:
+        logger.info(f"STDERR:\n{result.stderr}")
+
+    assert result.returncode == 0, f"openjd run failed with return code {result.returncode}"
+    assert "Session ended successfully" in result.stdout
```
crowecawcaw, line 141: Feels like we're missing some assertions. What about asserting on the "Custom step script running inside UE" log line?

crowecawcaw:
The PR title and description make it sound like we're adding an E2E test but the actual tests just looks like it verifies the bundle can start an adapter script inside UE. It doesn't use the submitter, a bundle from the submitter, or do a render than I can tell. Seems like a good unit test, but I'm not sure what testing gaps it will close. Could you share a bit about the functionality we're aiming to cover in this PR?

karthikbekalp (approved):

PR #321, opened 2026-06-22
fix: P4 workspace root change detection and dependency sync optimization

baxeaz:

Two performance and correctness issues with Perforce workspace syncing on render workers:

1. **Stale have-list when workspace Root changes** — When using persistent Perforce workspaces (`P4_CLIENTS_ROOT_DIRECTORY`), if the workspace Root changes between jobs (e.g. persistent volume recycled, directory moved), the P4 server's have-list still describes files at the old root. A subsequent `p4 sync` reports "file(s) up-to-date" and writes nothing to the new root, causing UE to fail loading assets.

2. **Redundant dependency sync on every job** — UE's `init_unreal.py` calls `SourceControl.sync_files()` with every job dependency individually on every launch. Even when all files are already at the correct revision, the P4 server round-trip takes significant time (scales with project size). On CMF, the sync environment template was not syncing dependencies at all (oversight — SMF already had it), so this was the only sync path. On reused workers, this was pure waste.

**Root change detection:**
- When `P4_CLIENTS_ROOT_DIRECTORY` is set, fetch the server-side workspace spec and compare Root to the new Root
- If they differ, run `p4 sync -k //...#none` to clear the stale have-list before saving the updated spec
- The subsequent sync then transfers files normally to the new root

**Dependency sync optimization:**
- Add `-MrqJobDependenciesDescriptor` to the CMF sync template (SMF already had it)
- Sync dependencies **without** `-f` (force) — P4's have-list skips files already at the correct revision, making reuse near-instant
- Emit `openjd_env: DEPENDENCIES_SYNCED=true` after successful dependency sync
- In `init_unreal.py`, skip the redundant `sync_files` call when `DEPENDENCIES_SYNCED` is set; still perform the asset registry scan so UE discovers the files

**Developer experience:**
- Log P4 sync results to clearly show which files were transferred vs skipped
- Warn when DeadlineWorker service needs restart after `--worker` install

- **Reused workspace, same changelist:** Eliminates redundant per-file P4 server round-trip during UE launch. The time saved depends on P4 server latency and dependency count — in our testing with MeerkatDemo (~250 assets), this eliminated ~30s of overhead per job.
- **Reused workspace, new changelist:** Only changed files transfer; others skip instantly
- **Root change:** Files correctly arrive at the new root (was broken without the flush)
- **First run / new root:** Full sync happens once in P4SyncCmf, UE skips redundant re-check
- Applies equally to CMF and SMF

- Unit tests for root-change detection (with/without registry), non-force dep sync, and DEPENDENCIES_SYNCED signal emission
- Manual E2E testing on Prodmon Windows CMF queue:
  - Root change with no local registry — flush fired, files synced correctly
  - Reuse with same changelist — all files "up-to-date" in ~2s, UE skipped sync
  - New changelist — only modified file (`Main_LVL.umap`) transferred, rest skipped
  - Fresh root — full sync in P4SyncCmf (~19s), UE skip confirmed
  - Backward compat — jobs using old template (without `-MrqJobDependenciesDescriptor`) fall through to UE's original sync

Yes — multiple successful render jobs on Prodmon Windows CMF queue across all scenarios above.

- Docstrings/comments updated for modified functions
- Log messages added for P4 sync visibility
- Template change is self-documenting (matches existing SMF template pattern)

No. Existing jobs without `-MrqJobDependenciesDescriptor` in the template fall through to UE's original `sync_files` behavior. Users must recreate/reset the P4SyncCmf data asset to pick up the new CMF template.

----

### DIFF
--- a/src/deadline/unreal_perforce_utils/app.py
+++ b/src/deadline/unreal_perforce_utils/app.py
@@ -92,6 +92,16 @@ class WorkspaceInfoFile:
         self.view_workspace = workspace_name
 
 
+def _normalize_root(root: str) -> str:
+    """
+    Normalize a workspace Root path for comparison: forward slashes, no trailing
+    separator, casefolded (P4 stores Windows paths case-insensitively).
+    """
+    if not root:
+        return ""
+    return root.replace("\\", "/").rstrip("/").casefold()
+
+
 def merge_view_mappings(existing_views: list[str], new_views: list[str]) -> list[str]:
     """
     Merge new view mappings into existing ones.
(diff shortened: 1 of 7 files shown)


src/deadline/unreal_perforce_utils/app.py
```diff
@@ -92,6 +92,16 @@ def register_view(self, workspace_name: str) -> None:
         self.view_workspace = workspace_name
 
 
+def _normalize_root(root: str) -> str:
+    """
+    Normalize a workspace Root path for comparison: forward slashes, no trailing
+    separator, casefolded (P4 stores Windows paths case-insensitively).
+    """
+    if not root:
+        return ""
```
crowecawcaw, line 101: Nit: don't think this is needed

crowecawcaw (dismissed):
Would be great to add perforce integ tests eventually

crowecawcaw (approved):

baxeaz:
> Would be great to add perforce integ tests eventually

Yeah agree - We'd need to have a server we keep around for it most likely, probably worth looking into for the additional coverage though

PR #337, opened 2026-07-11
feat(p4): SubmitMode for pushing render outputs into Perforce

baxeaz:
Fixes: *<insert link to GitHub issue here>*

P4-integrated render jobs could sync inputs from Perforce, but there was no way to push render outputs back into Perforce as part of the same job. Customers who use Perforce as their source of truth had to shuttle render results out of Job Attachments (or S3) into their depot as a separate manual step, which fragmented review workflows and made per-job provenance harder to track.

The requirement was a first-class, opt-in way to commit — or shelve, for review — the frames a Deadline Cloud render job produced, aggregated across every worker and every task into a single changelist.

Added a new job parameter `SubmitMode` on the P4 render job template with three values:

- `''` (default) — Job Attachments only, no P4 write. Preserves existing behavior.
- `submit` — Commit the aggregated outputs as one Perforce CL.
- `shelve` — Aggregate as before, but leave the final CL shelved for human review.

Under the hood:

1. Each render task shelves the files it produced under its output directory. The shelve is scoped to just that task's frames — the adaptor takes a pre/post-render mtime snapshot and passes the explicit file list to `p4 reconcile`, so chunked tasks sharing an `output_path` don't reshelve each other's frames.
2. Each per-task shelved CL description carries a marker line: `DeadlineCloudRenderShelve/<deadline-job-id>`.
3. A new `AssembleShelves` step runs once per job (after Render, via an OpenJD `dependsOn`), queries `p4 changes -s shelved -u <p4-user> -l`, filters by marker, and unshelves every match into a single aggregate CL. It then either submits or shelves that aggregate based on `SubmitMode`.
4. Cleanup of the per-task source shelves/CLs is best-effort — those temporary CLs are deleted after their content is folded in.

`revert -a` on the aggregate strips no-op file opens, so a rerender that produces bit-identical bytes doesn't inflate the CL. Existing P4 output frames that would be overwritten are handled by clearing the read-only bit before staging (previously-committed frames come off the workspace as read-only under `noallwrite`).

**Key assumption: all render workers for a given job authenticate to Perforce as the same P4 user.** `AssembleShelves` uses `-u <p4-user>` to keep the marker query cheap, so per-worker P4 identities would cause the aggregation step to miss shelves. This matches the existing recommended setup (single shared P4 secret in AWS Secrets Manager) documented in `perforce-credentials-management.md`.

- Opt-in: default `SubmitMode=''` is a strict no-op — no additional P4 traffic, no CL creation, no behavior change for existing jobs.
- Additive to Job Attachments — this does not replace JA upload. Enabling `SubmitMode` pushes the same outputs into Perforce in addition to whatever JA is configured to do.
- One CL per job regardless of task or worker count. Empty aggregates (nothing changed since last sync) fail loudly so operators aren't surprised by silence.

- Unit tests: 59 tests in `test/deadline_perforce_utils_for_unreal/test_app.py` (all passing), 40 tests in `test/deadline_adaptor_for_unreal/unit/UnrealAdaptor/test_adaptor.py` (all passing), 244 tests in the submitter suite (all passing). New tests cover marker-based shelve discovery, per-task explicit-files scoping, `unshelve -f` clobber semantics, split unshelve/cleanup client-switching, partial unshelve failure aggregation, post-shelve `revert -k` for downstream unshelve, and the `SubmitMode=''` no-op path.
- End-to-end on Deadline Cloud: verified on both single-worker and 2-worker fleet configurations. Single-worker end-to-end submit produced a 2,521-file CL; 2-worker end-to-end submit produced a correctly-merged CL from cross-client shelves.

Yes — multiple test iterations against the meerkat scene using SubmitMode and verifying JA flow still works

- CHANGELOG entry added under `## Unreleased`.
- New `Submitting Render Outputs Back to Perforce (SubmitMode)` section in `docs/user_guide/create-perforce-render-job.md`, including the marker convention and the "all workers share a P4 user" assumption.
- New `SubmitMode` row added to the parameter table in the same doc.

No. `SubmitMode` defaults to `''`, which is a no-op. Existing job templates and existing render jobs behave identically.

### DIFF
--- a/src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
+++ b/src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
@@ -487,6 +487,14 @@ class UnrealAdaptor(Adaptor[AdaptorConfiguration]):
             Action("set_handler", {"handler": run_data.get("handler", "base")})
         )
 
+        # Snapshot output_path contents BEFORE the render runs. Chunked tasks
+        # share one session (and therefore one output_path); the render only
+        # writes into the shot subdirs for *this* chunk, but the session dir
+        # also contains prior chunks' output. Without this snapshot we'd
+        # re-shelve every prior task's frames on every task. See the diff-
+        # against-snapshot logic in _maybe_submit_renders_to_perforce.
+        pre_render_snapshot = self._snapshot_output_files(run_data.get("output_path"))
+
         self._unreal_is_rendering = True
         self._action_queue.enqueue_action(Action("run_script", run_data))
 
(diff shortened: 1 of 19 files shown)


src/deadline/unreal_perforce_utils/app.py
```diff
@@ -521,132 +564,827 @@ def create_workspace(
...
+    Build a CL description that traces back to the Deadline session that
+    produced the renders. Reads the standard DEADLINE_* env vars the worker
+    agent injects into every job.
+
+    :param deadline_job_id: If provided, prepends a machine-parseable marker
+        line ``DeadlineCloudRenderShelve/<job-id>`` as the first line of the
+        description. AssembleShelves uses this to discover render shelves
+        belonging to a specific job across all workers.
     """
+    job_id = deadline_job_id or os.getenv("DEADLINE_JOB_ID")
```
Cherie-Chen: In src/deadline/unreal_perforce_utils/app.py, `aggregate_desc` specifically calls out to set `deadline_job_id=None` to handle resubmit workflow, but this fall back statement to get job ID from env will break this?
```
# Intentionally *omit* deadline_job_id from _build_changelist_description
# here so the aggregate CL's description does NOT carry the
# `DeadlineCloudRenderShelve/<job-id>` marker line. If it did, a retry
# of this step (or a re-submitted job with the same DEADLINE_JOB_ID)
# would discover the previously-created aggregate shelve via
# `_find_shelved_cls_for_job` and re-wrap it as if it were a task shelve.
aggregate_desc = _build_changelist_description(
    job_name=os.getenv("DEADLINE_JOB_ID"),
    extra_description=(...),
    deadline_job_id=None,
)
```

src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
```diff
@@ -515,6 +523,295 @@ def on_run(self, run_data: dict) -> None:
...
+        # push the outputs into the worker's Perforce client. This runs
+        # same-worker (we're inside the task's on_run, on the worker that
+        # produced the files), so OutputPath paths in run_data resolve to
+        # local files that exist.
+        #
+        # When SubmitMode is active, the submitter's
+        # `RenderUnrealOpenJob.get_asset_references` moves output_directories
+        # into referenced_paths, so Job Attachments no longer uploads render
+        # outputs — the P4 shelve here is the sole delivery path for the
+        # frames. Any failure inside `_maybe_submit_renders_to_perforce`
```
Cherie-Chen, line 536: Curious if you have consider doing P4 sync + uploading to JobAttachment instead of one or the other. I wonder how the download would work in the case where output is not uploaded to JobAttachment

src/deadline/unreal_perforce_utils/app.py
```diff
@@ -521,132 +564,827 @@ def create_workspace(
...
+        value = os.getenv(env_var)
+        if value:
+            lines.append(f"{label}: {value}")
+    if extra_description:
+        lines.append("")
+        lines.append(extra_description)
+    return "\n".join(lines)
+
+
+def edit_for_overwrite(unreal_project_name: str, depot_paths: list[str]) -> None:
```
Cherie-Chen: Looks like this is never used?

src/deadline/unreal_perforce_utils/app.py
baxeaz: Good catch — you're right, the `or os.getenv("DEADLINE_JOB_ID")` fallback made `deadline_job_id=None` collapse to the same job ID the task shelves used, so the aggregate CL was still getting the task-shelve marker. My "fix" wasn't a fix.

Actually fixed now in the latest commit:
- Removed the env-var fallback from `_build_changelist_description` — marker is stamped only when the caller explicitly passes a job ID.
- Added a distinct `DEADLINE_CL_AGGREGATE_MARKER_PREFIX = "DeadlineCloudRenderAggregate/"`.
- `assemble_shelves` stamps the aggregate CL with that prefix, so it's invisible to `_find_shelved_cls_for_job` which filters on the task prefix only.

Verified on a real fleet: the aggregate CL description now reads `DeadlineCloudRenderAggregate/job-...` as its first line (not `DeadlineCloudRenderShelve/...`). Also added three unit tests: (a) no env-var fallback for the marker, (b) aggregate marker gets stamped correctly, (c) `_find_shelved_cls_for_job` returns only task shelves even when an aggregate shares the job ID.

src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
baxeaz: Good question. The reasoning for making them mutually exclusive was:

1. **Cost**: uploading render outputs to S3 via JA when they're already going into P4 means paying the S3 storage bill for a duplicate copy, plus doubling the post-render upload time on the critical path.
2. **Source of truth**: for teams that adopt this feature, P4 becomes the artifact system of record for outputs. Having both paths risks confusion about which copy is canonical.
3. **Path mapping preserved**: the submitter moves the output directories from `asset_references.output_directories` into `referenced_paths`. That keeps them known to Deadline for path-mapping purposes (so downstream jobs and callers still resolve them correctly), but suppresses the S3 upload specifically.

That said, you're right that the download side is a real consideration — anyone consuming outputs through the Deadline Monitor, or a downstream job that doesn't have a P4 client, won't find them in S3. That was an intentional tradeoff for teams that use P4 as their delivery mechanism (they'd have downstream steps that `p4 sync` the outputs), but it's not universally the right call.

A few ways we could evolve this:
- **Add a `SubmitMode=both` value** so customers who want redundancy can pay the cost explicitly.
- **Leave S3 upload on by default even in `submit`/`shelve` modes** and let customers opt out via a separate parameter — safer default, but every current SubmitMode user pays the double cost.
- **Ship as-is** with the exclusivity documented, and add `both` in a follow-up when someone actually asks for it.

Happy to go with any of these. Want me to add `both` in this PR, or defer until we see the ask?

src/deadline/unreal_perforce_utils/app.py
baxeaz: Right, unused. `edit_for_overwrite` was from an earlier iteration where I thought I'd need to open files for edit before the staging copy to clear the P4-set read-only bit. The final design does that via `dst.chmod(stat.S_IWRITE | stat.S_IREAD)` in the adaptor's staging loop, and content-level differencing via `p4 reconcile -e -a` in `submit_renders` — so this function ended up with no callers. Deleted the function and its four unit tests in the latest commit.

src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
```diff
@@ -515,6 +523,295 @@ def on_run(self, run_data: dict) -> None:
...
+                staged += 1
+            except Exception as e:
+                logger.warning(
+                    "Failed to stage %r -> %r for Perforce submit: %s. Skipping this file.",
+                    str(src),
+                    str(dst),
+                    e,
+                )
+
+        if not workspace_paths_for_reconcile:
```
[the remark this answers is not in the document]
baxeaz: Real bug — good catch. Fixed in the latest commit.

Changed the per-file staging loop from log-and-continue to collect-and-raise-once:
- Per-file failures still log at WARNING (so operators get individual detail in the task log).
- After the loop, if `staging_failures` is non-empty, we raise a `RuntimeError` that includes the count (`N of M`), an inline preview of the first 5 failing (src -> dst: error) triples, and a "...and K more" tail if there are additional failures. Enough for postmortem without spamming a 100-line traceback when a whole batch fails.
- Any single failed copy now fails the task — matches the strictness of the other SubmitMode guards.

Tests: renamed `test_raises_when_all_copy2_fail_so_zero_files_staged` → `test_raises_when_all_copy2_fail` and added `test_raises_on_partial_copy2_failure` that stages 2 files where the second raises, and asserts (a) RuntimeError with `1 of 2 task-produced file(s) failed to stage`, (b) `submit_renders` is never called (no incomplete CL produced), (c) the per-file WARNING still lands.

src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
Cherie-Chen: Agree that we can defer the `both` option until we hear from customer, but could you please call out that the download feature will not work for `submit`/`shelve` options?

src/deadline/unreal_perforce_utils/app.py
```diff
@@ -521,132 +564,798 @@ def create_workspace(
...
+
+    # Step 2: create a numbered changelist with the description.
+    cl_description = _build_changelist_description(
+        job_name=os.getenv("DEADLINE_JOB_ID"),
+        extra_description=description,
+        deadline_job_id=deadline_job_id,
+    )
+    change_spec = connection.p4.fetch_change()
+    change_spec["Description"] = cl_description
+    saved = connection.p4.save_change(change_spec)
```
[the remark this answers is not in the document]
baxeaz: Belt-and-suspenders fix applied. Added `change_spec["Files"] = []` before `save_change` in both `submit_renders` and `assemble_shelves`, so the new numbered CL is guaranteed to start empty regardless of what `fetch_change` decides to auto-populate. The subsequent scoped `reopen -c <cl> <reconcile_targets>` then brings in only this task's files (for submit) or the unshelve fills the aggregate (for assemble).

Two new test assertions pin the clear so a future refactor can't regress it:
- `test_submit_renders_reconciles_and_submits_when_diffs_exist`: `saved_change.get("Files") == []`
- `test_assemble_shelves_discovers_and_aggregates_by_job_marker`: same check on the aggregate CL

Worth noting: on my fleet runs to date the numbered CLs have landed at the expected file counts, so if the auto-populate is happening in this flow the default CL happens to be empty at fetch time (probably because the pre-reconcile `revert -k` cleared us and no other job is co-tenant on this workspace). But that's a fragile invariant — this fix makes the correctness argument independent of what's currently in default.

Cherie-Chen (approved):

kavmur (approved):

PR #353, opened 2026-07-28
feat: add OpenJD TASK_CHUNKING dynamic chunking support

Cherie-Chen:
feat: add OpenJD TASK_CHUNKING dynamic chunking support

Deadline Cloud’s existing Unreal Engine render partitioning modes determine task boundaries when a job is submitted. This prevents the scheduler from dynamically grouping frames based on current worker availability and target runtime, limiting load balancing for workloads whose per-frame render times vary.

This change adds an opt-in third partitioning mode using OpenJD’s native `TASK_CHUNKING` extension and `CHUNK[INT]` task parameter type. Deadline Cloud determines contiguous frame chunks at dispatch time from the submitted frame range.

Added dedicated dynamic-chunking job and step templates that:

- Declare the OpenJD `TASK_CHUNKING` extension.
- Define a `CHUNK[INT]` task parameter.
- Expose `Frames`, `TargetRuntimeSeconds`, and `RangeConstraint` job parameters.
- Require contiguous ranges because Unreal Engine’s Movie Render Queue cannot render non-contiguous frame lists.

Updated the submitter to:

- Detect and validate dynamically chunked steps.
- Populate `Frames` from the MRQ frame range, converting MRQ’s exclusive end frame to an inclusive OpenJD range.
- Support negative frame numbers.
- Reject missing or empty frame ranges, non-contiguous constraints, and multiple `CHUNK[INT]` parameters in one step.
- Skip fixed `TaskIndex` range generation when the scheduler will form the chunks.
- Handle `CHUNK[INT]` consistently when Unreal Data Assets cannot represent that OpenJD parameter type.

Updated the adaptor to:

- Parse scheduler-provided contiguous frame ranges.
- Convert OpenJD’s inclusive end frame to MRQ’s exclusive end frame.
- Use the chunk’s starting frame for `{task_index}` filename substitution, preventing output collisions when dynamic chunks do not have a task index.

The render partitioning design and migration plan were also updated, including diagrams for all three partitioning modes.

Dynamic chunking is opt-in through the dedicated templates under `openjd_templates/dynamic_chunking`. The default submission path remains unchanged and continues to use the existing shots-per-task behavior.

Existing jobs and templates are unaffected. The new `dynamic_chunked_frames` run-data property is optional.

The dynamic templates require an adaptor release containing this new parsing and MRQ configuration logic. The templates specify `unrealengine-openjd=0.7.*` because this feature is intended for a 0.7.x patch release; the currently released `0.7.0` package predates this feature and cannot execute these templates successfully.

- Full unit suite: `677 passed, 2 skipped`
- Dynamic validator suite: `63 passed`
- E2E test ran with all tests passing in CMF setup

Yes, rendered test jobs in CMF

Yes, but I am also working a new user guide which will be merged in a different PR

No.

The feature is additive and opt-in. Existing templates, submitted jobs, and the default submission path retain their current behavior. The new adaptor run-data property is optional.

Users selecting the dedicated dynamic-chunking templates must use an adaptor release containing this feature; released `unrealengine-openjd=0.7.0` does not yet include the required adaptor logic.

----

### DIFF
--- a/src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
+++ b/src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
@@ -99,6 +99,9 @@ class UnrealOpenJobParameterDefinition:
         build_kwargs = dict(name=u_param.name, type=u_param.type.name)
         if u_param.value:
             python_class = PARAMETER_DEFINITION_MAPPING[u_param.type.name].python_class
+            # Unreal's ValueType enum never maps to task-only types such as CHUNK[INT],
+            # so a python_class is always available here.
+            assert python_class is not None
             build_kwargs["value"] = python_class(u_param.value)
         return cls(**build_kwargs)
 
@@ -380,13 +383,23 @@ class UnrealOpenJob(UnrealOpenJobEntity):
         :rtype: JobTemplate
         """
 
+        parameter_definitions = []
+        for param in self.get_template_object()["parameterDefinitions"]:
+            job_parameter_class = PARAMETER_DEFINITION_MAPPING[
+                param["type"]
+            ].job_parameter_openjd_class
+            if job_parameter_class is None:
+                # Task-only types (e.g. CHUNK[INT]) have no job-level equivalent
+                raise exceptions.SubmitterInputValidationError(
+                    f'Job parameter "{param.get("name")}" has type "{param["type"]}" '
+                    f"which is not valid at the job level"
+                )
+            parameter_definitions.append(job_parameter_class(**param))
+
         template_dict = {
             "specificationVersion": settings.JOB_TEMPLATE_VERSION,
             "name": self.name,
-            "parameterDefinitions": [
-                PARAMETER_DEFINITION_MAPPING[param["type"]].job_parameter_openjd_class(**param)
-                for param in self.get_template_object()["parameterDefinitions"]
-            ],
+            "parameterDefinitions": parameter_definitions,
             "steps": [s.build_template() for s in self._steps],
         }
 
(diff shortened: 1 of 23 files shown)


src/deadline/unreal_submitter/unreal_open_job/unreal_open_job_dynamic_chunking.py
```diff
@@ -0,0 +1,214 @@
...
+        # pattern (r"^(-?\d+)-(-?\d+)$") so submitter and adaptor accept the same
+        # expressions.
+        single_int = r"-?\d+"
+        range_with_optional_step = r"-?\d+--?\d+(?::\d+)?"
+        int_range_element_pattern = rf"(?:{range_with_optional_step}|{single_int})"
+        frames_validation_pattern = re.compile(
+            rf"^{int_range_element_pattern}(?:,{int_range_element_pattern})*$"
+        )
+
+        if not frames_validation_pattern.match(frames):
```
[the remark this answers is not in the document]
Cherie-Chen: The bot’s assumption is incorrect.

With `rangeConstraint: CONTIGUOUS`, OpenJD guarantees that dispatched chunk values are canonical `<start>-<end>` ranges—including singleton chunks such as `5-5`. The [TASK_CHUNKING RFC](https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/rfcs/0001-task-chunking.md) explicitly demonstrates:

- Input Frames: 1,10-12,18-50
- Dispatched chunks: 1-1, 10-12, 18-27, etc.
- CONTIGUOUS always expands to ranges such as 1-1 or 11-20

Therefore:

- The submitter correctly accepts OpenJD input expressions such as bare integers, lists, and stepped ranges.
- The scheduler converts those values into contiguous dispatched chunks.
- The adaptor correctly accepts only the scheduler-facing `<start>-<end>` representation.
- Restricting `Frames` to a single range would unnecessarily remove valid OpenJD pickup-frame and stepped-range support.

But, the code comment claiming the submitter and adaptor “accept the same expressions” is misleading. I will fix that

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
```diff
@@ -1148,6 +1161,116 @@ def _build_parameter_values_for_p4(self, parameter_values: list[dict]) -> list[d
...
+        if not frames_param:
+            return parameter_values
+
+        # Skip if Frames already has a value
+        if frames_param.get("value") is not None:
+            return parameter_values
+
+        # Need MRQ job to extract frame range
+        if not self._mrq_job:
+            logger.warning(
```
kavmur, line 1215: Any reason these are warnings and not validation errors? same for level seq.

kavmur (approved):

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
Cherie-Chen: Good catch—there isn’t an intentional reason. These started as defensive fallback paths, but reaching either branch means the required Frames parameter is unset and cannot be derived. Returning here leaves it unresolved and only defers the failure downstream.

I will cut a follow PR to fix this

PR #371, opened 2026-08-18
feat: add Unreal profiling output support

kavmur:
Fixes: N/A

Unreal render jobs could not request Unreal Insights, CSV Profiler, or MemReport capture through the Deadline submitter. Profiling files must be finalized at the correct render-task boundary and registered as job outputs so users can retrieve complete artifacts through the standard Deadline output-download path.

- Added opt-in Movie Render Queue settings for Unreal Insights CPU, GPU, and memory tracing.
- Added opt-in CSV Profiler capture with a configurable frame count.
- Added opt-in MemReport capture.
- Preserved explicit user command-line values over profiling preset defaults.
- Added render lifecycle handling that starts, stops, and finalizes task profiling before reporting task completion.
- Added a startup Insights trace for project initialization, including shader compilation, while retaining a separate trace for each render task.
- Enabled process-level memory instrumentation when Memory Insights is requested so memory events are available in startup and task traces.
- Finalized Insights traces into unique per-task files, including retries for transient Windows file locks and path handling for environment variables and whitespace.
- Registered profiling directories beside the configured Movie Render Queue output directory as Job Attachments outputs, including for Perforce SubmitMode jobs.
- Added the profiling fields to the adaptor run-data schema.

Users can enable any combination of Unreal Insights, CSV Profiler, and MemReport capture from render job settings and retrieve the generated artifacts as Deadline job outputs. Profiling options remain disabled by default, so non-profiling jobs retain their existing behavior. Explicit user command-line profiling arguments continue to take precedence over preset-derived defaults.

- Ran `hatch run lint` successfully.
- Ran the full Python test suite successfully: 773 passed, 2 skipped.
- Built and installed the plugin with Unreal Engine 5.7.
- Ran profiling-disabled and all-profiling worker end-to-end jobs on a customer-managed fleet.
- Verified the disabled job produced no profiling files.
- Downloaded a new 892 MB task trace, 70 KB CSV profile, and 510 KB MemReport through Job Attachments, then validated the Insights header and memory metadata, non-empty CSV rows, and complete MemReport begin/end markers.

Screenshot 2026-08-18 at 10 43 57 AM

Yes. Profiling-disabled and all-profiling render jobs completed successfully on a customer-managed fleet. The disabled job produced no profiling outputs, while the enabled job uploaded complete Unreal Insights, CSV Profiler, and MemReport artifacts as Deadline job outputs.

Yes. The submit-job use-case documentation and submitter setup guide describe the profiling controls, generated output locations, and Memory Insights instrumentation behavior. The adaptor run-data schema also documents the task-level profiling fields.

No. The new profiling controls and run-data fields are optional and disabled by default. No existing required adaptor inputs or public interfaces were removed or renamed.

----

### DIFF
--- a/src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
+++ b/src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
@@ -67,6 +67,10 @@ class UnrealAdaptor(Adaptor[AdaptorConfiguration]):
         super().__init__(*args, **kwargs)
 
         self.data_validation = DataValidation()
+        self._csv_capture_frames: Optional[int] = None
+        self._memreport_enabled: bool = False
+        self._insights_categories: Optional[str] = None
+        self._startup_insights_trace_file: Optional[str] = None
 
     @property
     def integration_data_interface_version(self) -> SemanticVersion:
@@ -270,6 +274,12 @@ class UnrealAdaptor(Adaptor[AdaptorConfiguration]):
         :param match: re.Match object from the regex pattern that was matched the message
         :type match: re.Match
         """
+        if match.groups() and (
+            self._csv_capture_frames or self._memreport_enabled or self._insights_categories
+        ):
+            logger.info("Waiting for profiling cleanup before completing the render task")
+            return
+
         self._unreal_is_rendering = False
         self.update_status(progress=100)
 
@@ -342,7 +352,6 @@ class UnrealAdaptor(Adaptor[AdaptorConfiguration]):
 
         # Remove the -execcmds argument from the extra_cmd_args
         extra_cmd_str = re.sub(r'(-execcmds=["\'][^"\']*["\'])', "", extra_cmd_str)
-
         client_path = self.unreal_client_path.replace("\\", "/")
         log_args = ["-log", "-unattended", "-stdout", "-allowstdoutlogverbosity", "-nozen"]
 
@@ -351,10 +360,34 @@ class UnrealAdaptor(Adaptor[AdaptorConfiguration]):
             log_args += ["-NoLoadingScreen", "-NoScreenMessages", "-RenderOffscreen", "-nozen"]
 
         extra_cmd_args = extra_cmd_str.split(" ")
+        extra_cmd_args, self._csv_capture_frames = self._extract_csv_capture_frames_arg(
+            extra_cmd_args
+        )
+        extra_cmd_args, self._memreport_enabled = self._extract_memreport_arg(extra_cmd_args)
+        extra_cmd_args, self._insights_categories = self._extract_insights_arg(extra_cmd_args)
+        if self._insights_categories and any(
+            re.match(r"^-trace(?:=|$)", arg, flags=re.IGNORECASE) for arg in extra_cmd_args
+        ):
+            logger.warning(
+                "Raw -trace arguments take precedence over Deadline Cloud Insights profiling"
+            )
+            self._insights_categories = None
+        trace_file_arg = self._get_tracefile_arg(unreal_project_path, " ".join(extra_cmd_args))
+        if self._insights_categories:
+            self._startup_insights_trace_file = self._get_startup_trace_file(unreal_project_path)
 
         args = [unreal_exe, unreal_project_path]
         args.extend(log_args)
         args.extend(extra_cmd_args)
+        if trace_file_arg:
+            args.append(trace_file_arg)
+        if self._startup_insights_trace_file:
+            args.extend(
+                [
+                    f"-trace={self._insights_categories}",
+                    f"-tracefile={self._startup_insights_trace_file}",
+                ]
+            )
         args = [arg for arg in args if arg]  # Remove empty strings
         args = list(dict.fromkeys(args))  # Remove duplicates
 
@@ -382,6 +415,150 @@ class UnrealAdaptor(Adaptor[AdaptorConfiguration]):
             stdout_handler=regexhandler,
             stderr_handler=regexhandler,
         )
+
+    @staticmethod
+    def _extract_csv_capture_frames_arg(
+        extra_cmd_args: list[str],
+    ) -> tuple[list[str], Optional[int]]:
+        """
+        Remove launch-time CSV frame capture so render steps can start CSV only
+        after MRQ reports that rendering has actually begun.
+        """
+
+        filtered_args: list[str] = []
+        csv_capture_frames: Optional[int] = None
+        index = 0
+
+        while index < len(extra_cmd_args):
+            token = extra_cmd_args[index]
+            match = re.match(r"^-csvCaptureFrames(?:=(.+))?$", token, flags=re.IGNORECASE)
+            if not match:
+                filtered_args.append(token)
+                index += 1
+                continue
+
+            raw_value = match.group(1)
+            if raw_value is None:
+                next_index = index + 1
+                if next_index < len(extra_cmd_args) and not str(
+                    extra_cmd_args[next_index]
+                ).startswith("-"):
+                    raw_value = extra_cmd_args[next_index]
+                    index = next_index
+
+            if raw_value is None or str(raw_value).startswith("-"):
+                logger.warning("Ignoring -csvCaptureFrames without a numeric value")
+                index += 1
+                continue
+
+            try:
+                csv_capture_frames = max(1, int(str(raw_value).strip()))
+                logger.info(
+                    "Deferring CSV capture until render begins: %s frame(s)",
+                    csv_capture_frames,
+                )
+            except ValueError:
+                logger.warning("Ignoring invalid -csvCaptureFrames value: %s", raw_value)
+
+            index += 1
+
+        return filtered_args, csv_capture_frames
+
+    @staticmethod
+    def _extract_memreport_arg(extra_cmd_args: list[str]) -> tuple[list[str], bool]:
+        """
+        Remove the MemReport transport flag from launch args so the render step
+        can request MemReport -full after MRQ has completed.
+        """
+
+        filtered_args: list[str] = []
+        memreport_enabled = False
+
+        for token in extra_cmd_args:
+            if re.match(r"^-MemReport(?:=.*)?$", token, flags=re.IGNORECASE):
+                if not memreport_enabled:
+                    logger.info("Deferring MemReport generation until render completion")
+                memreport_enabled = True
+                continue
+
+            filtered_args.append(token)
+
+        return filtered_args, memreport_enabled
+
+    @staticmethod
+    def _extract_insights_arg(extra_cmd_args: list[str]) -> tuple[list[str], Optional[str]]:
+        filtered_args: list[str] = []
+        insights_categories: Optional[str] = None
+
+        for token in extra_cmd_args:
+            match = re.match(r"^-DeadlineCloudInsights=(.+)$", token, flags=re.IGNORECASE)
+            if not match:
+                filtered_args.append(token)
+                continue
+
+            raw_value = match.group(1).strip().strip("\"'")
+            categories = [category.strip() for category in raw_value.split(",") if category.strip()]
+            if categories and all(
+                re.fullmatch(r"[A-Za-z0-9_.-]+", category) for category in categories
+            ):
+                insights_categories = ",".join(dict.fromkeys(categories))
+                logger.info(
+                    "Configuring Deadline Cloud Unreal Insights capture: %s",
+                    insights_categories,
+                )
+            else:
+                logger.warning("Ignoring invalid -DeadlineCloudInsights value: %s", raw_value)
+
+        return filtered_args, insights_categories
+
+    @staticmethod
+    def _get_or_create_trace_directory(unreal_project_path: str) -> Optional[str]:
+        if not unreal_project_path:
+            return None
+
+        project_dir = os.path.dirname(unreal_project_path).replace("\\", "/").rstrip("/")
+        trace_dir = f"{project_dir}/Saved/Profiling/DeadlineCloud"
+        try:
+            os.makedirs(trace_dir, exist_ok=True)
+        except OSError as exc:
+            logger.warning(
+                "Unable to create Unreal Insights trace directory %s: %s", trace_dir, exc
+            )
+            return None
+
+        return trace_dir
+
+    @staticmethod
+    def _get_tracefile_arg(unreal_project_path: str, extra_cmd_str: str) -> Optional[str]:
+        """
+        If tracing is enabled without an explicit tracefile, write traces under the
+        runtime project's Saved/Profiling directory so artifacts are produced on the worker.
+        """
+
+        if not unreal_project_path:
+            return None
+
+        if not re.search(r"(^|\s)-trace=", extra_cmd_str, flags=re.IGNORECASE):
+            return None
+
+        if re.search(r"(^|\s)-tracefile=", extra_cmd_str, flags=re.IGNORECASE):
+            return None
+
+        trace_dir = UnrealAdaptor._get_or_create_trace_directory(unreal_project_path)
+        if not trace_dir:
+            return None
+
+        trace_name = time.strftime("deadline-cloud-insights-%Y%m%d-%H%M%S.utrace")
+        return f"-tracefile={trace_dir}/{trace_name}"
+
+    @staticmethod
+    def _get_startup_trace_file(unreal_project_path: str) -> Optional[str]:
+        trace_dir = UnrealAdaptor._get_or_create_trace_directory(unreal_project_path)
+        if not trace_dir:
+            return None
+
+        trace_name = time.strftime("deadline-cloud-insights-startup-%Y%m%d-%H%M%S.utrace")
+        return f"{trace_dir}/{trace_name}"
 
     def _populate_client_loaded_action(self) -> None:
         """
@@ -481,6 +658,19 @@ class UnrealAdaptor(Adaptor[AdaptorConfiguration]):
             self.data_validation.validate_run_data(run_data)
         except (jsonschema.exceptions.ValidationError, jsonschema.exceptions.SchemaError) as e:
             self._record_error_and_raise(exc=e, exception_scope="on_run")
+
+        if run_data.get("handler", "base") == "render":
+            if self._csv_capture_frames or self._memreport_enabled or self._insights_categories:
+                run_data = dict(run_data)
+                if self._csv_capture_frames:
+                    run_data["csv_capture_frames"] = self._csv_capture_frames
+                if self._memreport_enabled:
+                    run_data["memreport"] = True
+                if self._insights_categories:
+                    run_data["insights_categories"] = self._insights_categories
+                    if self._startup_insights_trace_file:
+                        run_data["startup_insights_trace_file"] = self._startup_insights_trace_file
+                        self._startup_insights_trace_file = None
 
         # Set up the step handler
         self._action_queue.enqueue_action(
--- a/src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
+++ b/src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
@@ -8,6 +8,7 @@ import sys
 import time
 import logging
 import threading
+import uuid
 import jsonschema
 from typing import Callable, Optional
 
@@ -67,6 +68,9 @@ class UnrealAdaptor(Adaptor[AdaptorConfiguration]):
         super().__init__(*args, **kwargs)
 
         self.data_validation = DataValidation()
+        self._csv_capture_frames: Optional[int] = None
+        self._memreport_enabled: bool = False
+        self._insights_categories: Optional[str] = None
 
     @property
     def integration_data_interface_version(self) -> SemanticVersion:
@@ -342,7 +346,6 @@ class UnrealAdaptor(Adaptor[AdaptorConfiguration]):
 
         # Remove the -execcmds argument from the extra_cmd_args
         extra_cmd_str = re.sub(r'(-execcmds=["\'][^"\']*["\'])', "", extra_cmd_str)
-
         client_path = self.unreal_client_path.replace("\\", "/")
         log_args = ["-log", "-unattended", "-stdout", "-allowstdoutlogverbosity", "-nozen"]
 
@@ -351,10 +354,26 @@ class UnrealAdaptor(Adaptor[AdaptorConfiguration]):
             log_args += ["-NoLoadingScreen", "-NoScreenMessages", "-RenderOffscreen", "-nozen"]
 
         extra_cmd_args = extra_cmd_str.split(" ")
+        extra_cmd_args, self._csv_capture_frames = self._extract_csv_capture_frames_arg(
+            extra_cmd_args
+        )
+        extra_cmd_args, self._memreport_enabled = self._extract_memreport_arg(extra_cmd_args)
+        extra_cmd_args, self._insights_categories = self._extract_insights_arg(extra_cmd_args)
+        if self._insights_categories and any(
+            re.match(r"^-trace(?:=|$)", arg, flags=re.IGNORECASE) for arg in extra_cmd_args
+        ):
+            logger.warning(
+                "Raw -trace arguments take precedence over task-scoped Deadline Cloud "
+                "Insights profiling"
+            )
+            self._insights_categories = None
+        trace_file_arg = self._get_tracefile_arg(unreal_project_path, " ".join(extra_cmd_args))
 
         args = [unreal_exe, unreal_project_path]
         args.extend(log_args)
         args.extend(extra_cmd_args)
+        if trace_file_arg:
+            args.append(trace_file_arg)
         args = [arg for arg in args if arg]  # Remove empty strings
         args = list(dict.fromkeys(args))  # Remove duplicates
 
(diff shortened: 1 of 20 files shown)


src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
```diff
@@ -351,10 +354,31 @@ def _start_unreal_client(self) -> None:
...
+            category.lower() == "memory" for category in self._insights_categories.split(",")
+        )
 
         args = [unreal_exe, unreal_project_path]
         args.extend(log_args)
         args.extend(extra_cmd_args)
+        if trace_file_arg:
+            args.append(trace_file_arg)
+        if memory_insights_enabled:
+            args.append("-trace=memory")
```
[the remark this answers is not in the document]
kavmur: The ordering is intentional so -trace=memory initializes memory tracing without creating a launch-scoped output file.

src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
```diff
@@ -351,10 +354,31 @@ def _start_unreal_client(self) -> None:
...
+        extra_cmd_args, self._memreport_enabled = self._extract_memreport_arg(extra_cmd_args)
+        extra_cmd_args, self._insights_categories = self._extract_insights_arg(extra_cmd_args)
+        if self._insights_categories and any(
+            re.match(r"^-trace(?:=|$)", arg, flags=re.IGNORECASE) for arg in extra_cmd_args
+        ):
+            logger.warning(
+                "Raw -trace arguments take precedence over task-scoped Deadline Cloud "
+                "Insights profiling"
+            )
+            self._insights_categories = None
```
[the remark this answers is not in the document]
kavmur: CMF E2E produced a finalized task .utrace containing memory events, confirming Trace.File works.

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
```diff
@@ -21,12 +23,275 @@
...
+            )
+            executor.insightsFinished = True
+            return
+
+    trace_filename = os.path.basename(executor.insightsTraceFile)
+    working_trace_file = os.path.join(tempfile.gettempdir(), trace_filename)
+    command_trace_file = working_trace_file
+    if any(character.isspace() for character in command_trace_file):
+        command_trace_file = trace_filename
+        working_trace_file = unreal.Paths.convert_relative_path_to_full(trace_filename)
```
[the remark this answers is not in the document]
kavmur: Low risk, only happens if the workers temp directory contains spaces.

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
```diff
@@ -21,12 +23,278 @@
...
+            if attempt < 49:
+                time.sleep(0.1)
+
+    logger.warning(
+        "Render Executor: Unable to move finalized Insights trace from '%s' to '%s': %s",
+        source,
+        destination,
+        last_error,
+    )
+    return False
```
phil-IO-p: The temp-directory staging in `_start_executor_insights_trace` plus the `shutil.move` here worries me a little for large traces (the PR description mentions an 892 MB one!).

Three concerns I'd appreciate your opinion on:

1. Claude noted that on Linux workers `tempfile.gettempdir()` is usually a different mount from the project directory, so `shutil.move` degrades to copy-then-delete rather than a rename. On the larger end, that could be a full ~900 MB copy on the render-completion path, per task, and if Unreal hasn't finished flushing when the copy starts we'd copy a partial file and then unlink the source. 
2. The retry budget is `range(50)` with `time.sleep(0.1)`, so 5 s total. Is that enough time to finish writing and release the handle on a trace that size after `Trace.Stop`? For a small trace it seems fine; for 892 MB I'm less sure.
3. When the move does fail we `logger.warning` and return `False`, but `_stop_executor_insights_trace` still sets `insightsFinished = True` and `_handle_executor_render_completion` still logs `"Render Executor: Rendering is complete"`. So the task succeeds, the trace is missing from the output directory, and the only evidence is a warning in the worker log. That's a little concerning.

It seems the staging is there because `Trace.File` can't take a path containing spaces, is that right? If so, would it work to write directly to the final directory when that path has no whitespace (the branch you already have using `unreal.Paths.profiling_dir()` shows the shape), and keep the temp hop only for the whitespace case? That removes the cross-device copy in the common case. Separately, would it be worth treating a failed finalize as something the user can actually see, given they paid the render cost for it?

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
```diff
@@ -155,10 +515,7 @@ def regex_pattern_complete() -> list[re.Pattern]:
         :return: A list of regular expression patterns
         :rtype: list[re.Pattern]
         """
-        return [
-            re.compile(".*Render Executor: Rendering is complete"),
-            re.compile(".* finished ([0-9]+) jobs in .*"),
-        ]
+        return [re.compile(".*Render Executor: Rendering is complete")]
```
phil-IO-p: Dropping `.* finished ([0-9]+) jobs in .*` makes sense; however, it also removes the only completion signal that didn't depend on the Python `on_executor_finished_delegate` being delivered. Post-change, a
render that genuinely finishes only ever completes via `_handle_executor_render_completion`, and that function isn't wrapped in a try/except at the top level, so an exception on the first `executor.renderCompletionHandled` read would leave `adaptor.py`'s `while self._unreal_is_rendering` loop spinning to the task timeout.

Claude checked the failure path and that one seems fine: `executor_failed_callback` still logs
`Render Executor: Error:`, which `regex_pattern_error` matches, and `_has_exception` raises. So this is specifically about the scenario where "render succeeded but the delegate didn't fire".

Since `renderCompletionHandled` already makes the profiling teardown idempotent, would keeping
the MRQ pattern as a backstop cost anything? A late second match would just re-run
`_handle_complete`, which only sets a flag and updates progress.

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
```diff
@@ -155,10 +515,7 @@ def regex_pattern_complete() -> list[re.Pattern]:
         :return: A list of regular expression patterns
         :rtype: list[re.Pattern]
         """
-        return [
-            re.compile(".*Render Executor: Rendering is complete"),
-            re.compile(".* finished ([0-9]+) jobs in .*"),
```
leon-li-inspire, line 160: What's the reason we removed this complete pattern ? Could you elaborate what could be the behavior modifications ?

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
```diff
@@ -132,6 +132,128 @@ def to_dict(self):
...
+        profiling_root = f"{project_saved_directory}/Profiling"
+        if self.is_insights_enabled():
+            return [profiling_root]
+
+        output_directories = []
+        if self.csv_profiler:
+            output_directories.append(f"{profiling_root}/CSV")
+        if self.memreport:
+            output_directories.append(f"{profiling_root}/MemReports")
+        return output_directories
```
phil-IO-p: The Insights branch returns the whole `Saved/Profiling` root while the CSV and MemReport branches return the specific leaves. Since `_get_task_trace_file` writes traces specifically to
`{project}/Saved/Profiling/DeadlineCloud` (and your E2E test asserts the `.utrace` shows up
under `/deadlinecloud/`), would returning `f"{profiling_root}/DeadlineCloud"` work here and
keep all three branches consistent?

Please correct me if I'm wrong but `Saved/Profiling` is Unreal's general profiling dump directory, so the current approach could result in anything else sitting there getting registered as this job's output too. That would have a bigger impact on workers with a persistent workspace, where the directory can survive between jobs.

src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
```diff
@@ -383,6 +407,148 @@ def _start_unreal_client(self) -> None:
...
+            os.makedirs(trace_dir, exist_ok=True)
+        except OSError as exc:
+            logger.warning(
+                "Unable to create Unreal Insights trace directory %s: %s", trace_dir, exc
+            )
+            return None
+
+        task_label = run_data.get("task_index", run_data.get("chunk_id", "task"))
+        trace_name = f"deadline-cloud-insights-task-{task_label}-{uuid.uuid4().hex}.utrace"
+        return f"{trace_dir}/{trace_name}"
```
phil-IO-p, line 561: This gives each task its own uuid-suffixed filename in `Saved/Profiling/DeadlineCloud` and Claude couldn't find anything that removes a finalized trace after the task's outputs are uploaded. Since one adaptor session serves many tasks (your existing comment in `on_run` notes chunked tasks share a session), traces look like they accumulate for the life of the session. At the ~892 MB from the PR description that's roughly 9 GB after ten tasks.

One concerning failure mode is a full disk partway through a long chunked job, which would surface as an unrelated-looking render failure on task N rather than as a profiling problem.

Would it be reasonable to remove the previous task's trace at the top of `_get_task_trace_file`, or sweep the directory at task start? It's not clear to me that the adaptor has a clean signal for "upload finished", which is why sweeping at the start of the *next* task might be the easier hook.

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
```diff
@@ -180,6 +537,9 @@ def executor_failed_callback(executor, pipeline, is_fatal, error):
         :param is_fatal: Whether the error is fatal or not
         :param error: The error message
         """
+        executor = executor or UnrealRenderStepHandler.active_executor
+        UnrealRenderStepHandler._stop_executor_csv_capture(executor, "render error")
+        UnrealRenderStepHandler._stop_executor_insights_trace(executor, "render error")
         logger.error(f"Render Executor: Error: {error}")
```
phil-IO-p, line 619: As far as I can tell a failed task doesn't run the output upload, so the just-finalized artifacts stay on the worker and are lost. Is that intended?

Two smaller things Claude found in the same function:

- `_generate_executor_memreport` isn't called on this path, so a failed render produces no MemReport. Is that deliberate? A memory report after a failure seems like one of the more useful cases, especially for an out-of-memory crash.
- It doesn't set `renderCompletionHandled`, so `wait_result`'s branch that clears `active_executor` never runs and the class attribute keeps referencing the dead executor until the next `run_script`. Harmless today, but see my separate note about the three teardown paths.

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
```diff
@@ -1659,6 +1811,11 @@ def _get_mrq_job_attachments_output_directories(self) -> list[str]:
 
         return output_directories
 
+    def _get_profiling_output_directories(self) -> list[str]:
+        return self.profiling_settings.get_output_directories(
+            unreal.SystemLibrary.get_project_saved_directory()
+        )
```
phil-IO-p, line 1832: This feeds `unreal.SystemLibrary.get_project_saved_directory()` straight into `f"{project_saved_directory}/Profiling"` with only a `.replace("\\", "/").rstrip("/")`. The neighboring methods in this class normalize first: `_get_mrq_job_attachments_output_directories` maps each entry through
`common.os_abs_from_relative(...)`, and `_get_mrq_job_output_directory` runs `validate_path_does_not_contain_non_valid_chars`.

Claude flagged this because the only other use of that API in the repo (`unreal_open_job_step.py:786`)
passes its result through `unreal.Paths.convert_relative_path_to_full`, i.e. that call site doesn't assume it's already absolute. Do you know whether it returns an absolute path or Unreal's `../../../MyProject/Saved/` form? If the latter, it seems like we'd be registering a relative output directory and the path-mapping rule wouldn't resolve on the worker.

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
```diff
@@ -21,12 +23,278 @@
...
+            "stop Insights trace",
+        ),
+    ):
+        try:
+            operation()
+        except Exception:
+            logger.exception("Render Executor: Failed to %s", description)
+
+    # Completion must not depend on optional profiling operations succeeding.
+    logger.info("Render Executor: Rendering is complete")
```
phil-IO-p: Flagging an inconsistency here. There are three places that signal "the render is over, tear down profiling" and each does a slightly different subset of operations:

- This one: all three operations, sets `renderCompletionHandled` and logs the completion line
- `UnrealRenderStepHandler._handle_executor_completion` (fallback branch): same three operations and the same log line, but doesn't set `renderCompletionHandled`
- `executor_failed_callback`: stop CSV and stop Insights only (no MemReport), doesn't set the flag either.

Since `wait_result` clears `active_executor` only on `renderCompletionHandled`, 2 of the 3 paths leave it set. That's benign right now because the adaptor completes off the stdout regex rather than the flag, but the duplication means the next change to teardown has to be made in three places to stay consistent.

Would it be worth having one function own teardown + flag + log, with the error path calling it with a different reason string? That'd also force an explicit decision on whether MemReport should run after a failed render.

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
```diff
@@ -132,6 +132,128 @@ def to_dict(self):
...
+                    ),
+                    300,
+                ),
+            ),
+            memreport=cls._coerce_bool(
+                cls._read_unreal_property(
+                    profiling_settings, "memreport", "mem_report", "bMemReport"
+                )
+            ),
+        )
```
phil-IO-p, line 227: `_read_unreal_property` tries up to three spellings per field with a `try/except Exception: continue`
around the `get_editor_property` attempts. Since the Unreal-Python name for `bInsightsCpu` is
`insights_cpu`, the first candidate always hits and the other two are unreachable today.

Regarding the failure behavior... if the C++ field is ever renamed, every candidate misses, `_read_unreal_property` returns `None`, `_coerce_bool` defaults to `False`, and the result is a user who ticks the profiling boxes, gets a successful job, and gets no profiling and no error. The sibling
`JobSharedSettings.from_u_deadline_cloud_job_shared_settings` reads its fields directly
(`initial_state=job_shared_settings.initial_state`), so a rename there raises at submit time, which is the louder and likely more useful outcome. What do you think?

Would reading the fields directly work here too? If a fallback is genuinely needed for older plugin binaries, keeping one alternate spelling and logging a warning when it's used would at least make the degraded case visible.

phil-IO-p (changes requested):

test/deadline_adaptor_for_unreal/unit/UnrealClient/step_handlers/test_render_step_handler.py
```diff
@@ -147,3 +153,613 @@ def test_no_partitioning_keys_is_noop(self, unreal_render_step_handler):
...
+
+        with patch(
+            "deadline.unreal_adaptor.UnrealClient.step_handlers.unreal_render_step_handler."
+            "UnrealRenderStepHandler._handle_executor_completion"
+        ) as handle_completion_mock:
+            unreal_render_step_handler.wait_result()
+
+        handle_completion_mock.assert_not_called()
+        assert type(unreal_render_step_handler).active_executor is executor
+        assert type(unreal_render_step_handler).render_wait_started is True
```
phil-IO-p: `..._while_rendering`, and `..._when_render_temporarily_stops` differ only in `executor.renderStarted` and `executor.is_rendering.return_value`, but the new `wait_result` reads neither of those. It looks at `handler_cls.render_wait_started`, `handler_cls.active_executor`, and `getattr(executor, "renderCompletionHandled", False)`. So all three end up on the same fall-through branch with the same assertions.

Why I brought this up... 3 test names describing 3 render states reads like the wait behavior is well covered, when one test covers it. If `renderStarted` / `is_rendering()` are meant to matter here, maybe `wait_result` should consult them (the fall-through branch currently can't distinguish "not started yet" from "in progress"). If they're not, collapsing these into one test of the keep-waiting branch
would be clearer

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
phil-IO-p: In `_handle_executor_render_completion` the 3 console commands are issued and then
`"Render Executor: Rendering is complete"` is logged straight away, which is what ends the task and lets the worker upload outputs. The trace has the 50-attempt move loop to wait for Unreal, but `CsvProfile Stop` and `MemReport -full` have no wait.

Do you know whether Unreal flushes those two synchronously inside the console command? I couldn't
determine it from the code here. If either is deferred, the upload could win the race and the user gets a truncated or missing file.

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
kavmur: 1. "finished N jobs in " is emitted by unreal engine when it finishes the job in executor queue, then unreal fires OnExecutorFinishedDelegate where our plugin callback runs.
2. " Rendering is complete" is emitted by our plugin AFTER we stop/finalize profiling insights

If unreal emits this log "finished ... jobs" BEFORE our profiling completes and cleaned up, the adaptor marks the task complete and uploads the outputs, this may cause a race and some of the profiling output could be incomplete or missing.
So the behavior before: task is marked completed/outputs uploaded if we receive either "Rendering is complete" OR "finished N jobs"..
After this change: task is marked completed only when we receive "Rendering is complete"

kavmur: The chances that the delegate fails and the message from our callback never fires is low, but I could also make this behavior change only affect if profiling is enabled.

src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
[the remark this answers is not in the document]
kavmur: Addressed. Deadline-managed Insights now launches with an explicit startup `-tracefile`, which is finalized into the profiling outputs before the task trace starts.

src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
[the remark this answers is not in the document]
kavmur: Addressed. The startup trace is stopped and finalized before `Trace.File` starts the task trace. If startup finalization fails, the task trace is skipped rather than starting a second session.

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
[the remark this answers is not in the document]
kavmur: Addressed. Relative trace filenames are now resolved against `unreal.Paths.profiling_dir()`, with unit coverage for paths containing whitespace.

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
```diff
@@ -21,12 +23,278 @@
...
+        (lambda: _generate_executor_memreport(executor), "generate MemReport"),
+        (
+            lambda: _stop_executor_insights_trace(executor, "render completed"),
+            "stop Insights trace",
+        ),
+    ):
+        try:
+            operation()
+        except Exception:
+            logger.exception("Render Executor: Failed to %s", description)
```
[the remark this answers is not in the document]
kavmur: Addressed. Profiling cleanup exceptions are logged as warnings, and unit coverage verifies they cannot suppress successful render completion.

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
```diff
@@ -1659,6 +1755,11 @@ def _get_mrq_job_attachments_output_directories(self) -> list[str]:
 
         return output_directories
 
+    def _get_profiling_output_directories(self) -> list[str]:
+        return self.profiling_settings.get_output_directories(
+            unreal.SystemLibrary.get_project_saved_directory()
```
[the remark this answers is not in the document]
kavmur: Addressed. Profiling outputs are now derived from the configured MRQ output directory rather than `get_project_saved_directory()`.

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
kavmur: Updated based on the feedback: non-profiling behavior is unchanged. Only profiling jobs ignore the earlier engine completion message and wait for profiling finalization.

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
kavmur: Addressed. Insights registers only `Profiling/DeadlineCloud`; CSV and MemReport register their corresponding leaf directories.

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
kavmur: Addressed. The profiling path now follows the same resolved MRQ output-directory path used for render outputs and creates a sibling `Profiling` directory.

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
kavmur: Addressed. The engine `finished N jobs` pattern remains a completion fallback for non-profiling jobs. For profiling jobs, the adaptor waits for the explicit plugin completion message emitted after profiling cleanup.

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
kavmur: Addressed for the feature path. The current executor owns successful profiling teardown, idempotency, and completion. The remaining branch is a compatibility fallback for executors without the new hook; the render-error path intentionally remains separate.

test/deadline_adaptor_for_unreal/unit/UnrealClient/step_handlers/test_render_step_handler.py
kavmur: Addressed. The equivalent tests were collapsed into one test covering the actual keep-waiting behavior.

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
kavmur: Addressed. The plugin now exposes explicit CSV-write and MemReport completion checks. Task completion waits for both, with a 300-second warning-only timeout.

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
kavmur: Addressed. Alternate Unreal property spellings remain supported, but if every candidate fails the submitter now raises `SubmitterInputValidationError` instead of silently disabling profiling.

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
kavmur: This is a valid limitation because Deadline plugin does not upload outputs for failed tasks, while marking the task successful would hide the render failure. We will track failed-task diagnostic artifact upload as a follow-up.

src/deadline/unreal_adaptor/UnrealAdaptor/adaptor.py
kavmur: Correct but low risk. Profiling is opt-in and typically limited to sampled tasks, so we are treating this storage as profiling output cost. Also, there is no reliable client-side signal that upload has completed.

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
kavmur: Low risk. Supported workers are Windows and normally keep temporary and project data on the same volume; the 892 MB CMF trace finalized successfully. Cross-volume staging remains a low-risk edge case, and failures stay warnings so profiling cannot fail a successful render.

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
```diff
@@ -821,6 +956,13 @@ def mrq_job(self, value):
             if override_description:
                 self._description = override_description
 
+        if self._mrq_job is not None and self._mrq_job.preset_overrides is not None:
+            profiling_settings = getattr(self._mrq_job.preset_overrides, "profiling_settings", None)
+            if profiling_settings is not None:
```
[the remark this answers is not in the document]
kavmur: This can affect queues saved before the profiling fields existed, and reselecting the preset refreshes the overrides. Again low risk, profiling can materially slow Unreal rendering, it is intended for opt-in diagnostics rather than as a persistent preset default.

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job.py
```diff
@@ -1721,6 +1895,9 @@ def get_asset_references(self) -> AssetReferences:
             # Render output path
             asset_references.output_directories.add(self._get_mrq_job_output_directory())
 
+        profiling_output_directories = set(self._get_profiling_output_directories())
```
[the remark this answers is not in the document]
kavmur: _get_profiling_output_directories() now uses unreal.Paths.profiling_dir()

leon-li-inspire (approved):

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
```diff
@@ -20,13 +22,352 @@
...
+            if attempt < 49:
+                time.sleep(0.1)
+
+    logger.warning(
+        "Render Executor: Insights trace did not finalize at '%s': %s",
+        trace_file,
+        last_error,
+    )
+    return False
+
```
phil-IO-p, line 238: Writing the trace straight to its final path is a big improvement, thanks for handling that!                

Regarding the `try` block here... Claude pointed out this works on Windows, where UE's trace writer holds the file with a deny-write share mode, so the open fails until it's released. On Linux and macOS there's no mandatory locking, so opening a file another process still has open succeeds immediately. On those platforms the loop returns on the first attempt and the wait is effectively removed.                                                                                           
                                                                                                               
2 reasons it feels realistic not just theoretical: `openjd_templates/host_requirements.yml:16` is `anyOf: [windows, linux, macos]`, so Linux workers are in scope, and the E2E and release workflows are both `os: windows`, so nothing in CI would surface it.                                                              
                                                                                                               
Since you already added `IsCsvCaptureComplete` to the new C++ library, would a matching trace-writer-idle query fit there too? That'd be platform-independent and consistent with how CSV and MemReport completion now work. A size-stable-across-N-polls check would at least behave the same on both platforms.

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
```diff
@@ -20,13 +22,352 @@
...
+                stop_csv_capture()
+                executor.csvCompletionPending = True
+                logger.info(f"Render Executor: Requested CSV capture stop ({reason})")
+            except Exception as exc:
+                logger.warning("Render Executor: Unable to stop CSV capture: %s", exc)
+        elif _execute_editor_console_command("CsvProfile Stop"):
+            logger.info(f"Render Executor: Stopped CSV capture ({reason})")
+
+    executor.csvStarted = False
+    executor.csvFinished = True
```
phil-IO-p, line 168: On the exception path nothing stops the capture, and the function still sets `csvFinished = True` just below it, so there's no later retry. Since stopping the profiler is what flushes the `.csv`, that path would produce a missing or truncated file, right?
                                                                                                               
Would falling through to `_execute_editor_console_command("CsvProfile Stop")` inside the `except` block be reasonable? That keeps the old behavior as a safety net for the case where the new path fails.

kavmur: Both impact and risk of occurrence is low here.
Native C++ profiling calls should rarely throw unless there is legitimate unreal profiler issue. 
In the event of a failure, the rendering remains unaffected, the profiling outputs may be truncated.

src/deadline/unreal_adaptor/UnrealClient/step_handlers/unreal_render_step_handler.py
kavmur: This is a valid race, but again low risk for supported workflows, as our plugin currently only supports windows workloads. The plugin will not be available via conda for linux/mac workers

phil-IO-p (approved):
Chatted offline. Recent comments will be addressed in a follow-up as future improvements. Approved!

PR #396, opened 2026-09-18
fix: skip unpopulated shot-based MRQ jobs

kavmur:
Follow-up to #391.

Shot-based MRQ jobs with an unpopulated shot list could still create an invalid empty OpenJD task range.

Allow empty shot lists only for frame-based or dynamic-chunked jobs, and validate zero enabled shots during template construction.

Shot-based jobs are skipped before submission; valid frame-based and dynamic-chunked jobs remain eligible.

Hatch unit suite: 799 passed, 2 skipped. Hatch lint: Ruff, Black, and mypy passed.

No. This requires an Unreal editor and Deadline Cloud environment.

No. This behavior fix needs no documentation change.

No.

----


kavmur:
Superseded by #397, which includes this follow-up together with the #391 E2E regression fix.

PR #397, opened 2026-09-18
fix: validate MRQ submission eligibility

kavmur:
Fixes the MRQ submission regression and a followup from #391 

Empty shot lists could create an invalid OpenJD task range.

Allow empty lists only for frame-based or dynamic-chunked jobs, reject zero enabled shots during template construction, and guard MRQ eligibility inspection.

Invalid shot-based jobs are skipped and cannot create an empty task range.

`hatch run lint` passed. `hatch run test`: 801 passed, 2 skipped. E2E passed.

Yes, the earlier PR-targeted E2E run passed; 

No.

No.

----


src/unreal_plugin/Content/Python/remote_executor.py
```diff
@@ -57,6 +61,18 @@ def execute_delayed(self, pipeline_queue):
 
         unreal_submitter.submit_jobs()
 
+    @staticmethod
+    def _is_frame_based_job(job):
+        parameters = job.get_parameter_definition_with_overrides().parameters
```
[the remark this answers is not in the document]
kavmur: Fixed in 0aa48ad: jobs without Deadline parameter definitions are now treated as ineligible, with regression coverage.

src/unreal_plugin/Content/Python/remote_executor.py
```diff
@@ -57,6 +61,18 @@ def execute_delayed(self, pipeline_queue):
...
+        frames_per_task = next(
+            (parameter for parameter in parameters if parameter.name == "FramesPerTask"), None
+        )
+        return (
+            frames_per_task is not None
```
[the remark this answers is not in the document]
kavmur: Fixed in 0aa48ad: malformed FramesPerTask values are treated as non-frame-based, with regression coverage.

src/unreal_plugin/Content/Python/remote_executor.py
```diff
@@ -34,7 +34,11 @@ def execute_delayed(self, pipeline_queue):
             job
             for job in pipeline_queue.get_jobs()
             # An empty shot list may be unpopulated for programmatically-created jobs.
-            if job.enabled and (not job.shot_info or any(shot.enabled for shot in job.shot_info))
+            if job.is_enabled()
+            and (
+                any(shot.enabled for shot in job.shot_info)
+                or (not job.shot_info and self._is_frame_based_job(job))
+            )
```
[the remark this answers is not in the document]
kavmur: Keeping this PR focused on correctness; per-job skip diagnostics will be handled separately.

src/unreal_plugin/Content/Python/remote_executor.py
```diff
@@ -57,6 +61,18 @@ def execute_delayed(self, pipeline_queue):
...
+    @staticmethod
+    def _is_frame_based_job(job):
+        parameters = job.get_parameter_definition_with_overrides().parameters
+        frames_per_task = next(
+            (parameter for parameter in parameters if parameter.name == "FramesPerTask"), None
+        )
+        return (
+            frames_per_task is not None
+            and int(frames_per_task.value or 0) > 0
+            or any(parameter.name == "Frames" for parameter in parameters)
```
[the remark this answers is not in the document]
kavmur: Keeping this PR focused; parameter constants and expression cleanup will be handled separately.

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job_step.py
```diff
@@ -629,6 +629,12 @@ def _get_task_count(self) -> int:
         if shots_per_task <= 0:
             shots_per_task = 1  # by default each task renders 1 shot
 
+        if not enabled_shots:
+            raise exceptions.SubmitterInputValidationError(
+                "MRQ job has no enabled shots. Populate its shot list or set FramesPerTask to "
```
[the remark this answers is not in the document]
kavmur: Keeping this PR focused; the documentation and message cleanup will be handled separately.

src/unreal_plugin/Content/Python/remote_executor.py
```diff
@@ -34,7 +34,11 @@ def execute_delayed(self, pipeline_queue):
             job
             for job in pipeline_queue.get_jobs()
             # An empty shot list may be unpopulated for programmatically-created jobs.
-            if job.enabled and (not job.shot_info or any(shot.enabled for shot in job.shot_info))
+            if job.is_enabled()
+            and (
+                any(shot.enabled for shot in job.shot_info)
```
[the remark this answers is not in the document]
kavmur: Keeping this filter: frame-based worker tasks preserve disabled shots, so submitting an all-disabled job can produce no render output.

src/unreal_plugin/Content/Python/remote_executor.py
```diff
@@ -57,6 +61,25 @@ def execute_delayed(self, pipeline_queue):
...
+            (parameter for parameter in parameters if parameter.name == "FramesPerTask"), None
+        )
+        try:
+            has_frames_per_task = (
+                frames_per_task is not None and int(frames_per_task.value or 0) > 0
+            )
+        except (TypeError, ValueError):
+            has_frames_per_task = False
+
+        return has_frames_per_task or any(parameter.name == "Frames" for parameter in parameters)
```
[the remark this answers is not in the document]
kavmur: - Likelihood: Low — requires a custom template with that ambiguous shape.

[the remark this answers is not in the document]
kavmur: Fixed in 387eefa: dynamic eligibility now checks step-template CHUNK[INT] parameters rather than the Frames job parameter.

src/unreal_plugin/Content/Python/remote_executor.py
```diff
@@ -57,6 +61,25 @@ def execute_delayed(self, pipeline_queue):
 
         unreal_submitter.submit_jobs()
 
+    @staticmethod
+    def _is_frame_based_job(job):
+        get_parameters = getattr(job, "get_parameter_definition_with_overrides", None)
+        if get_parameters is None:
+            return False
+
+        parameters = get_parameters().parameters
```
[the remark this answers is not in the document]
kavmur: Fixed in 387eefa: the executor skips jobs without a preset, and the native parameter getter now returns an empty result safely.

src/unreal_plugin/Content/Python/remote_executor.py
```diff
@@ -57,6 +64,34 @@ def execute_delayed(self, pipeline_queue):
...
+        try:
+            has_frames_per_task = (
+                frames_per_task is not None and int(frames_per_task.value or 0) > 0
+            )
+        except (TypeError, ValueError):
+            has_frames_per_task = False
+
+        return has_frames_per_task or any(
+            DynamicChunkingHelper.is_using_dynamic_chunking_from_file(
+                step.path_to_template.file_path
```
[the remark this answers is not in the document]
kavmur: Deferred as pre-existing preset-data hardening outside this eligibility fix; we will address null step entries separately.

src/unreal_plugin/Source/UnrealDeadlineCloudService/Private/MovieRenderPipeline/MoviePipelineDeadlineCloudExecutorJob.cpp
```diff
@@ -247,6 +247,12 @@ FDeadlineCloudJobPresetStruct UMoviePipelineDeadlineCloudExecutorJob::GetDeadlin
 
 FDeadlineCloudJobParametersArray UMoviePipelineDeadlineCloudExecutorJob::GetParameterDefinitionWithOverrides() const
 {
+	if (!IsValid(JobPreset))
```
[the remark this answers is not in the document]
kavmur: Deferred as pre-existing sibling-path hardening outside this submission fix; we will add the matching guard separately.

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job_dynamic_chunking.py
```diff
@@ -57,6 +60,21 @@ def is_using_dynamic_chunking(step_template_object: dict) -> bool:
...
+
+        :param template_path: Path to a step template file
+        :return: True if the template has a CHUNK task parameter, False otherwise
+        :rtype: bool
+        """
+        if not template_path or not os.path.exists(template_path):
+            return False
+
+        with open(template_path, "r") as template_file:
+            return DynamicChunkingHelper.is_using_dynamic_chunking(yaml.safe_load(template_file))
```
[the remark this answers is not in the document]
kavmur: Fixed in ce04f18: unreadable or malformed templates now log a warning and are treated as non-dynamic, with regression coverage.

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job_dynamic_chunking.py
```diff
@@ -57,6 +60,29 @@ def is_using_dynamic_chunking(step_template_object: dict) -> bool:
...
+        :return: True if the template has a CHUNK task parameter, False otherwise
+        :rtype: bool
+        """
+        if not template_path or not os.path.exists(template_path):
+            return False
+
+        try:
+            with open(template_path, "r") as template_file:
+                template_object = yaml.safe_load(template_file)
+        except (OSError, yaml.YAMLError) as error:
```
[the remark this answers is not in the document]
kavmur: Likelihood: low; shipped templates and normal template-loading paths assume the platform text encoding already.

baxeaz: Confirmed defect: `UnicodeDecodeError` is neither `OSError` nor `yaml.YAMLError`, so it escapes this handler. Reproduced on PyYAML 6.0.3 with bytes that are invalid in the ambient encoding:

```
RAISED: UnicodeDecodeError | OSError? False | YAMLError? False | UnicodeError? True
```

PyYAML only rewraps decode failures as `ReaderError` when it performs the decoding itself (byte streams). For a text-mode handle `raw_decode` is `None`, so the error from `stream.read()` propagates unchanged, which defeats the documented intent that an unreadable template is simply not dynamically chunked.

Suggested:

```python
try:
    with open(template_path, "r", encoding="utf-8") as template_file:
        template_object = yaml.safe_load(template_file)
except (OSError, UnicodeError, yaml.YAMLError) as error:
    logger.warning(
        f"Could not read dynamic chunking step template {template_path}: {error}"
    )
    return False
```

Note that `encoding="utf-8"` on its own does **not** close this - invalid UTF-8 still raises. `UnicodeError` in the handler is the part that actually guards; the explicit encoding is for deterministic behavior across platforms, since the bare `open(path, "r")` currently decodes with `locale.getpreferredencoding()` (cp1252 on most Windows hosts).

For context, `unreal_open_job_entity.py:195` uses the same unguarded `open(path, "r")` + `safe_load` pattern, so this code follows existing convention; what's new here is the handler that claims to fail safe. Worth a test with invalid encoded bytes alongside the existing malformed-YAML one.

src/unreal_plugin/Content/Python/remote_executor.py
```diff
@@ -57,6 +64,34 @@ def execute_delayed(self, pipeline_queue):
...
+            )
+            for step in job_preset.steps
```
baxeaz, line 92: Confirmed defect. `Steps` is declared at `DeadlineCloudJob.h:251` as

```cpp
UPROPERTY(EditAnywhere, BlueprintReadWrite, Category = "Parameters", meta = (DisplayPriority = 6))
TArray<TObjectPtr<UDeadlineCloudStep>> Steps;
```

`EditAnywhere` on a `TObjectPtr` array lets a user add a row in the details panel and leave it unset, so Python sees `None` and `step.path_to_template` raises `AttributeError`.

I agree with the assessment that this isn't purely pre-existing hardening, and the distinction is concrete: the existing null-step traversals are `unreal_open_job.py:375` and `:1049`, both reached through `add_job` and so covered by its `error_notify` wrapper. This traversal is in the unwrapped queue-filter comprehension.

```python
for step in job_preset.steps
if step is not None
```

plus a regression test with `steps=[None]` asserting the job is ineligible rather than raising.

Related, very low severity: the `job_preset is None` check at line 70 is not equivalent to the new `IsValid(JobPreset)` guard in C++. A pending-kill preset is non-`None` in Python, passes line 70, and then `job_preset.steps` here touches a stale UObject.

src/unreal_plugin/Content/Python/remote_executor.py
```diff
@@ -34,7 +37,11 @@ def execute_delayed(self, pipeline_queue):
             job
             for job in pipeline_queue.get_jobs()
             # An empty shot list may be unpopulated for programmatically-created jobs.
-            if job.enabled and (not job.shot_info or any(shot.enabled for shot in job.shot_info))
+            if job.is_enabled()
+            and (
+                any(shot.enabled for shot in job.shot_info)
+                or (not job.shot_info and self._is_frame_based_job(job))
+            )
         ]
```
baxeaz, line 45: This comprehension is unwrapped, and that's what makes both reported defects worse than "the queue isn't processed."

Every other exit path in this method reaches `on_executor_finished_impl()`: line 33 (empty queue), line 48 (nothing eligible), line 118 in `check_dirty_packages`, line 133 in `check_maps`. Always signalling completion is an invariant here. An exception raised while evaluating this comprehension skips it, so MRQ never learns the executor finished and the editor sits in an in-progress state instead of surfacing an error.

Beyond the two specific guards, consider making that invariant unconditional, so a future traversal bug degrades to a skipped job rather than a hung editor:

```python
def _is_eligible(self, job):
    try:
        if not job.is_enabled():
            return False
        return any(shot.enabled for shot in job.shot_info) or (
            not job.shot_info and self._is_frame_based_job(job)
        )
    except Exception:
        logger.exception(
            f"Could not determine submission eligibility for job "
            f"`{getattr(job, 'job_name', '<unknown>')}`; skipping it."
        )
        return False

jobs_to_submit = [job for job in pipeline_queue.get_jobs() if self._is_eligible(job)]
```

Two smaller notes on this block:

- Line 79 hardcodes `"FramesPerTask"`, duplicating `OpenJobStepParameterNames.FRAMES_PER_TASK` (`unreal_open_job_entity.py:338`). This module already imports from `deadline.unreal_submitter.unreal_open_job`, so the constant is available.
- Line 40 uses `job.is_enabled()` while line 42 uses `shot.enabled`. The move away from `job.enabled` isn't explained in the diff, and the README supports UE 5.4 through 5.8 - a brief comment on the reason would prevent it from looking like an inconsistency.

src/deadline/unreal_submitter/unreal_open_job/unreal_open_job_dynamic_chunking.py
kavmur: Fixed in c0321dc: template reads now use UTF-8, catch UnicodeError, and cover invalid byte input.

src/unreal_plugin/Content/Python/remote_executor.py
kavmur: Fixed in c0321dc: unset step entries are skipped with regression coverage; pending-kill preset validation is deferred as low-risk follow-up.

src/unreal_plugin/Content/Python/remote_executor.py
kavmur: Deferred: a broad exception handler could hide eligibility errors; this PR keeps targeted guards, with broader completion and cleanup work for follow-up.

src/unreal_plugin/Content/Python/remote_executor.py
```diff
@@ -34,7 +37,11 @@ def execute_delayed(self, pipeline_queue):
             job
             for job in pipeline_queue.get_jobs()
             # An empty shot list may be unpopulated for programmatically-created jobs.
-            if job.enabled and (not job.shot_info or any(shot.enabled for shot in job.shot_info))
+            if job.is_enabled()
```
baxeaz, line 40: This is the one item I would want settled before merge, and it is the highest-consequence thing still open.

Commit `81ac7d12` ("fix: use MRQ job enabled accessor") deliberately changed `job.enabled` to `job.is_enabled()`. `README.md:23` declares support for **Unreal Engine 5.4 through 5.8**.

If `IsEnabled()` is not exposed to Python on the older engines in that range, this line raises `AttributeError` on *every* submission for those users. And because the comprehension it sits in is not wrapped (see my note at line 45), the failure skips `on_executor_finished_impl()` entirely -- so the symptom is a hung MRQ editor rather than an error message. That is a worse outcome than the bug this PR is fixing.

Could you confirm `is_enabled()` resolves on 5.4 and 5.5, or gate it, e.g.:

```python
is_enabled = job.is_enabled() if hasattr(job, "is_enabled") else job.enabled
```

If the accessor does exist across the whole supported range, disregard -- but a one-line comment saying why the switch was needed would save the next reader the same question.

src/unreal_plugin/Content/Python/remote_executor.py
```diff
@@ -34,7 +37,11 @@ def execute_delayed(self, pipeline_queue):
...
+            and (
+                any(shot.enabled for shot in job.shot_info)
```
baxeaz: Requesting this one-liner, because it is the direct counterpart to the null-step guard you just added and it is in the same expression.

`ShotInfo` is an array of `UMoviePipelineExecutorShot` **object pointers**, so unlike `Steps` -> `FFilePath`, a null entry here is genuinely representable. The comment on line 39 explicitly contemplates programmatically-created jobs, which is exactly the path that can leave an array slot unset.

So the current state is asymmetric: null steps are handled at line 93, null shots are not, and a null shot raises `AttributeError` on `shot.enabled` from an unwrapped comprehension -- the editor-hang path again.

```python
any(shot.enabled for shot in job.shot_info if shot is not None)
```

Worth a test case mirroring `null_step_job`. For reference, `unreal_open_job_step.py:587` has the same unguarded pattern, but that one runs inside `error_notify`, so it surfaces as a dialog instead of a hang -- lower priority, and arguably out of scope here.

src/unreal_plugin/Content/Python/remote_executor.py
baxeaz, line 45: Non-blocking, and weaker than I claimed last time -- keeping it on the record as insurance rather than as a fix.

Every other exit path in `execute_delayed` signals completion before returning: line 33 (empty queue), line 48 (nothing eligible), and inside `check_dirty_packages` and `check_maps`. "Always reach `on_executor_finished_impl()`" is an invariant in this file, and an exception escaping this comprehension is the one way to break it.

With the two concrete vectors now guarded, this no longer fixes a known crash. But it is what converts any *future* traversal bug -- including the `is_enabled()` scenario at line 40 if that turns out to be real -- from a silent editor hang into a logged error the user can see and report. Cheap for the protection it provides:

```python
try:
    jobs_to_submit = [...]
except Exception:
    logger.exception("Failed to determine which jobs are eligible for submission.")
    self.on_executor_finished_impl()
    return
```

Your call whether it belongs in this PR or a follow-up.

src/unreal_plugin/Content/Python/remote_executor.py
```diff
@@ -57,6 +64,35 @@ def execute_delayed(self, pipeline_queue):
...
+        if job_preset is None:
+            return False
+
+        get_parameters = getattr(job, "get_parameter_definition_with_overrides", None)
+        if get_parameters is None:
+            return False
+
+        parameters = get_parameters().parameters
+        frames_per_task = next(
+            (parameter for parameter in parameters if parameter.name == "FramesPerTask"), None
```
baxeaz, line 79: Non-blocking, carried over: `"FramesPerTask"` is hardcoded here but already exists as `OpenJobStepParameterNames.FRAMES_PER_TASK` (`unreal_open_job_entity.py:338`), and this file already imports from that package. Using the constant keeps this eligibility check from silently diverging if the parameter is ever renamed in the job template.

src/unreal_plugin/Content/Python/remote_executor.py
```diff
@@ -57,6 +64,35 @@ def execute_delayed(self, pipeline_queue):
...
+        )
+        try:
+            has_frames_per_task = (
+                frames_per_task is not None and int(frames_per_task.value or 0) > 0
+            )
+        except (TypeError, ValueError):
+            has_frames_per_task = False
+
+        return has_frames_per_task or any(
+            DynamicChunkingHelper.is_using_dynamic_chunking_from_file(
```
baxeaz, line 89: Non-blocking, carried over: this opens and re-parses the step template YAML for every step of every job on each submit, on the editor thread. With a large queue that is a noticeable stall in the UI before submission starts.

The sibling implementation `_is_using_dynamic_chunking` (`unreal_open_job_step.py:556`) already memoizes via `_dynamic_chunking_detected`. Either reuse that or add a small cache keyed on template path -- the templates cannot change mid-submit.

src/unreal_plugin/Content/Python/remote_executor.py
kavmur: Epic’s versioned Python API docs list MoviePipelineExecutorJob.is_enabled() in both UE 5.4 and 5.5.
- https://dev.epicgames.com/documentation/en-us/unreal-engine/python-api/class/MoviePipelineExecutorJob?application_version=5.4
- https://dev.epicgames.com/documentation/en-us/unreal-engine/python-api/class/MoviePipelineExecutorJob?application_version=5.5

src/unreal_plugin/Content/Python/remote_executor.py
kavmur: Fixed in 832643b: null shot entries are ignored, with coverage for the no-eligible-jobs path.

src/unreal_plugin/Content/Python/remote_executor.py
```diff
@@ -57,6 +64,35 @@ def execute_delayed(self, pipeline_queue):
...
+                step.path_to_template.file_path
+            )
+            for step in job_preset.steps
+            if isinstance(step, unreal.DeadlineCloudRenderStep)
```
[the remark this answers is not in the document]
kavmur: invalid, no-shot render cannot create tasks

src/unreal_plugin/Content/Python/remote_executor.py
```diff
@@ -34,7 +37,11 @@ def execute_delayed(self, pipeline_queue):
             job
             for job in pipeline_queue.get_jobs()
             # An empty shot list may be unpopulated for programmatically-created jobs.
-            if job.enabled and (not job.shot_info or any(shot.enabled for shot in job.shot_info))
+            if job.is_enabled()
+            and (
+                any(shot.enabled for shot in job.shot_info if shot is not None)
```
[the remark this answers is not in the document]
kavmur: low risk.  primary disabled/empty-shot regression is still fixed; such malformed jobs would fail later during submission rather than silently submit invalid work.

src/unreal_plugin/Content/Python/remote_executor.py
kavmur: I think the broad exception handling can conceal other real defects and logged error may not be really useful in troubleshooting.

src/unreal_plugin/Content/Python/remote_executor.py
kavmur: Valid maintenance cleanup (no impact) will do this as a followup in a future PR.

src/unreal_plugin/Content/Python/remote_executor.py
kavmur: low impact: after render-step scoping, it parses at most one render template per eligible no-shot job, not every preset step. Also a cache adds lifecycle/invalidation complexity

baxeaz (approved):

